PR #52, opened 2021-12-28
ordering

pifopi:
Add a program to order boxes of pokemon, based on their national dex order


Mysticial:
Can you give the program a better name? Like "National Dex Box-Reorder"? Just "Ordering" is kinda vague. lol

PR #59, opened 2022-01-09
Gift berry bot

Gin890:
- Fix compilation on macOS: size_t cannot be found unless a header like cstddef is included.
- Add berry name and sprite database, berry name reader, berry widget and berry table config.
- Add GiftBerryReset program to reset in front of the rare berry npc to get new types of berries.

### DIFF
--- a/SerialPrograms/Source/Pokemon/Resources/Pokemon_BerryNames.cpp
+++ b/SerialPrograms/Source/Pokemon/Resources/Pokemon_BerryNames.cpp
@@ -1,0 +1,114 @@
+/*  Pokemon Pokeball Names
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include <QtGlobal>
+#include <QString>
+#include "Common/Cpp/Exception.h"
+#include "Common/Qt/QtJsonTools.h"
+#include "CommonFramework/Globals.h"
+#include "Pokemon_BerryNames.h"
+
+namespace PokemonAutomation{
+namespace Pokemon{
+
+
+struct BerryNameDatabase{
+    BerryNameDatabase();
+
+    static const BerryNameDatabase& instance(){
+        static BerryNameDatabase database;
+        return database;
+    }
+
+    static const std::string NULL_SLUG;
+    std::vector<std::string> ordered_list;
+    std::map<std::string, BerryNames> database;
+    std::map<QString, std::string> reverse_lookup;
+};
+const std::string BerryNameDatabase::NULL_SLUG;
+
+// Currently only include berries used in BDSP.
+BerryNameDatabase::BerryNameDatabase()
+{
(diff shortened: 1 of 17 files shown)


SerialPrograms/Source/Pokemon/Resources/Pokemon_BerryNames.cpp
```diff
@@ -0,0 +1,203 @@
...
+        "starf-berry",
+        "enigma-berry",
+        "micle-berry",
+        "custap-berry",
+        "jaboca-berry",
+        "rowap-berry",
+        "roseli-berry",
+    }
+{
+    database["cheri-berry"  ].m_display_name = "Cheri Berry";
```
Mysticial, line 100: Can we pull this entire database from the BerryNameOCR file instead of hard-coding it here?

Mysticial: Seems like some of the other resources also abuse the OCR file for display names. The Pokemon name file is the only one that's has a "proper" display name file.

Let's just abuse the ENG section of the OCR file for these. So basically parse the file and collect all the slugs from there. Then iterate the English names for the display names.

Gin890: By reading the OCR json file, the internal order of the berries become alphabetical, because QJsonObject sorts its children objects when loading. I would prefer keeping its order as original but alphabetical order is also fine. Do you know if it's possible to traverse the children of QJsonObject using the order it is original written in the json file?

Mysticial: JSON doesn't define order. So that's not actually possible.

So the Pokeball list is hard-coded. And I'm not particularly happy about it. But that list is relatively short. Though teetering at the limit of where I'd make it a resource.

The Pokemon list is massive, but there are separate files for each dex list which is where the order comes from.

I feel like the proper thing here is:

1. A json for ordering.
2. A json for display names (translations).
3. A json for OCR. (or multiple jsons if the list is huge like the Pokemon list)

The Pokemon list is the only thing that does this right. IV Checker is very short and hard-coded in a ton of places already so that doesn't really count as a list. Pokeballs is wrong.

So how about this. Let's move berries to this format. Pokeballs will be refactored into it as well at a later time.

In the longer run, the display name json might be extended to include all items. While there might be separate lists for just berries, battle items, etc... which are ordered.

So I'm thinking:

- `ItemNameDisplay.json` - which will be all items following the same format as `PokemonNameDisplay.json`. But we can just put berries in there for now. If we ever need it, @denvoros is the most likely person to provide us a full item list. lol
- `ItemListBerries.json` - A single array of all of the berry slugs in the desired order. (similar to all the Pokedex lists)
- `BerryNameOCR.json` - Stays as is.

PR #61, opened 2022-01-18
Add auto catch

Gin890:
- Allow auto shiny overworld to catch pokemon automatically by calling basic_catcher()
- basic_catcher() is the same logic as SwSh basic_catcher()
- Plus can detect own fainted using arrow detector.
- And can handle learning new move by arrow detector and not learning them
- Tested on cases: wild pokemon caught, wild pokemon fainted, own pokemon fainted, learning new moves after gaining exp.
- Does not work on the situation that own pokemon evolves after battle or blacks out returning to Pokecenter.

### DIFF
--- a/SerialPrograms/Source/PokemonBDSP/Inference/Battles/PokemonBDSP_ExperienceGainDetector.cpp
+++ b/SerialPrograms/Source/PokemonBDSP/Inference/Battles/PokemonBDSP_ExperienceGainDetector.cpp
@@ -1,0 +1,79 @@
+/*  Experience Gain Detector
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include "Common/Compiler.h"
+#include "Common/Cpp/FixedLimitVector.tpp"
+#include "CommonFramework/Tools/VideoOverlaySet.h"
+#include "CommonFramework/ImageTools/SolidColorTest.h"
+#include "PokemonBDSP_ExperienceGainDetector.h"
+
+#include <iostream>
+using std::cout;
+using std::endl;
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonBDSP{
+
+
+ExperienceGainDetector::~ExperienceGainDetector(){}
+ExperienceGainDetector::ExperienceGainDetector(Color color)
+    : m_color(color)
+    , m_dialog(color)
+    , m_middle_column(0.365, 0.01, 0.008, 0.79)
+    , m_left_column(0.006, 0.01, 0.03, 0.79)
+    , m_lower_left_region(0.006, 0.8, 0.16, 0.15)
+{}
+
+void ExperienceGainDetector::make_overlays(VideoOverlaySet& items) const{
+    m_dialog.make_overlays(items);
+    items.add(m_color, m_middle_column);
+    items.add(m_color, m_left_column);
+    items.add(m_color, m_lower_left_region);
+}
+bool ExperienceGainDetector::detect(const QImage& screen) const{
+    if (!m_dialog.detect(screen)){
+        // std::cout << "ExperienceGainDetector: No dialogue detected, return" << std::endl;
+        return false;
+    }
+
+    const ImageStats stats0 = image_stats(extract_box(screen, m_middle_column));
+    if (!is_solid(stats0, {0.16, 0.42, 0.42}, 0.1, 30)){
+        // std::cout << "ExperienceGainDetector: No m_middle_column detected, " << stats0.average.to_string() << ", " << stats0.stddev.to_string() << std::endl;
+        return false;
+    }
+    const ImageStats stats1 = image_stats(extract_box(screen, m_left_column));
+    if (!is_solid(stats1, {0.3, 0.35, 0.35}, 0.1, 40)){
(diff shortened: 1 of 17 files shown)


SerialPrograms/Source/PokemonBDSP/Inference/Battles/PokemonBDSP_ExperienceGainDetector.cpp
```diff
@@ -0,0 +1,79 @@
...
+        return false;
+    }
+
+    const ImageStats stats0 = image_stats(extract_box(screen, m_middle_column));
+    if (!is_solid(stats0, {0.16, 0.42, 0.42}, 0.1, 30)){
+        // std::cout << "ExperienceGainDetector: No m_middle_column detected, " << stats0.average.to_string() << ", " << stats0.stddev.to_string() << std::endl;
+        return false;
+    }
+    const ImageStats stats1 = image_stats(extract_box(screen, m_left_column));
+    if (!is_solid(stats1, {0.3, 0.35, 0.35}, 0.03, 40)){
```
Mysticial, line 49: The 0.03 here (and below) seems really tight. Will likely break on Elgato and other "expensive cards". I'm guessing you don't have a 2 different cap cards of color shift to test this on right?

SerialPrograms/Source/PokemonBDSP/Inference/PokemonBDSP_ReceivePokemonDetector.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const ImageStats stats0 = image_stats(extract_box(frame, m_box0));
+    if (!is_solid(stats0, {0.22951, 0.340853, 0.429638}, 0.15, 20)){
+        return m_received;
+    }
+    const ImageStats stats1 = image_stats(extract_box(frame, m_box1));
+    if (!is_solid(stats1, {0.22951, 0.340853, 0.429638}, 0.15, 20)){
+        return m_received;
+    }
+    m_received = true;
+    return false;
```
Mysticial, line 45: This has the effect of not returning true until the blue screen is over right?

SerialPrograms/Source/PokemonBDSP/Programs/PokemonBDSP_BasicCatcher.cpp
```diff
@@ -0,0 +1,271 @@
...
+            results.result = CatchResult::POKEMON_CAUGHT;
+            break; //  Continue the loop.
+        case 2:
+            env.log("BasicCatcher: Detected move learn! Don't learn the new move.", COLOR_BLUE);
+            num_learned_moves++;
+            if (num_learned_moves == 100){
+                PA_THROW_StringException("BasicCatcher: Learn new move attempts reach 100.");
+            }
+            pbf_move_right_joystick(console, 128, 255, 20, 105);
+            pbf_press_button(console, BUTTON_A, 20, 105);
```
Mysticial, line 256: `BUTTON_A -> BUTTON_ZL`

SerialPrograms/Source/PokemonBDSP/Inference/PokemonBDSP_ReceivePokemonDetector.cpp
Gin890: Yes.

SerialPrograms/Source/PokemonBDSP/Inference/Battles/PokemonBDSP_ExperienceGainDetector.cpp
Gin890: I only have a MiraBox and a cheap Amazon capture card. Would them be enough?

Gin890: Updated the PR to change the max_distance_threshold from 0.03 to 0.1.

SerialPrograms/Source/PokemonBDSP/Programs/PokemonBDSP_BasicCatcher.cpp
Gin890: Updated the PR to address the comment.

SerialPrograms/Source/PokemonBDSP/Inference/Battles/PokemonBDSP_ExperienceGainDetector.cpp
Mysticial: Not really. lol

The problem here is that if we increase this to 0.15 like everything else, the high stddev tolerance (30 and 40) is enough to make this false positive on a lot of things.

Can the stddev tolerance be reduced back to the default of 10 without false positives?

Gin890: The regions I choose have a pretty high stddev. Default 10 is too small to catch them.
The stddevs are:
For my macOS screenshot, middle column 21, left column 24, lower left region 29.
For Switch native capture transferred via phone, middle column 26, left column 39, lower left region 25.
For Windows screenshot you sent me, middle column 29, left column 39, lower left region 27.

I notice when running away from a pokemon, the dialog box is wide. But when gaining exp, the dialog box is short. The exp gain detector uses the short dialog box detector to aid detection. It may be that the short dialog box is enough for detecting the exp gain. If this is the case, we can relax the color matching threshold without much worry about it gives false positives or false negatives.

PR #70, opened 2022-01-23
Add autonomous ball thrower for bdsp

pifopi:


Common/Microcontroller/DeviceRoutines.h
```diff
@@ -24,7 +24,7 @@ uint8_t program_id(const BotBaseContext& context);
 
 class DeviceRequest_seqnum_reset : public BotBaseRequest{
 public:
-    pabb_MsgInfoSeqnumReset params;
+    pabb_MsgInfoSeqnumReset params{};
```
Mysticial, line 27: What's going on here?

Mysticial: If it's the compiler complaining about a class having uninitialized fields, we should suppress that warning in the compile flags. We definitely do not want to zero-initialize every struct/class by default.

pifopi: Exactly. It was using a tool, I can remove those changes. I just thought it was clearer that way ^^

PR #71, opened 2022-01-23
Fix compile error and warning on macOS

Gin890:
- Fix unable to compile because of no definition on vector<QImage> in ExactImageDictionaryMatcher.cpp
- Fix compiler warning of no definition of ZERO_TILE and tile() in Kernels_SparseBinaryMatrixBase.h when compiling PokemonSwSh_MaxLair_Detect_PathSide.cpp.

### DIFF
--- a/SerialPrograms/Source/PokemonSwSh/MaxLair/Inference/PokemonSwSh_MaxLair_Detect_PathSide.cpp
+++ b/SerialPrograms/Source/PokemonSwSh/MaxLair/Inference/PokemonSwSh_MaxLair_Detect_PathSide.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <cmath>
+#include "Kernels/BinaryMatrix/Kernels_SparseBinaryMatrixBase.tpp"
 #include "Kernels/Waterfill/Kernels_Waterfill.h"
 #include "CommonFramework/ImageTools/ImageStats.h"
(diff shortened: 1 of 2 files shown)


SerialPrograms/Source/PokemonSwSh/MaxLair/Inference/PokemonSwSh_MaxLair_Detect_PathSide.cpp
```diff
@@ -5,6 +5,7 @@
  */
 
 #include <cmath>
+#include "Kernels/BinaryMatrix/Kernels_SparseBinaryMatrixBase.tpp"
```
Mysticial, line 8: Something is wrong here. The entire class is explicitly instantiated in `Kernels_BinaryMatrix.cpp`.

What's the error message you're getting?

pifopi: I guess it is the same that the warning I got there 
https://github.com/pifopi/Arduino-Source/runs/4914462996?check_suite_focus=true#step:8:370

Mysticial:
Can someone check if this is still needed after yesterday's code dump?

Gin890:
It's not needed any more. Close the PR. Thanks!

PR #81, opened 2022-02-21
Add test audio match program

Gin890:
First version of audio matching code in audio test program.


SerialPrograms/Source/PokemonLA/Inference/PokemonLA_ShinySoundDetector.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+
+ShinySoundDetector::ShinySoundDetector(Logger& logger)
+    : AudioInferenceCallback("ShinySoundDetector")
+    , m_logger(logger)
+{
+    // TODO : handle case with sample rate mismatch between file and capture card:
+
+    // TODO: move the template file to the resource folder
+    QString shinyFilename = "./heracrossShinyTemplateCompact.wav";
```
Mysticial, line 30: Let's go ahead and move this to the resource folder. Since we can't find an official asset from the data dump, let's just use this then.

SerialPrograms/Source/PokemonLA/Inference/PokemonLA_ShinySoundDetector.cpp
```diff
@@ -0,0 +1,78 @@
...
+namespace NintendoSwitch{
+namespace PokemonLA{
+
+
+
+ShinySoundDetector::ShinySoundDetector(Logger& logger)
+    : AudioInferenceCallback("ShinySoundDetector")
+    , m_logger(logger)
+{
+    // TODO : handle case with sample rate mismatch between file and capture card:
```
Mysticial, line 27: What happens if there is a match right now? Can you make it throw an error for now? We'll come back and add it later, but I don't want it to silently fail as is.

SerialPrograms/Source/CommonFramework/Inference/SpectrogramMatcher.cpp
```diff
@@ -0,0 +1,203 @@
...
+        return SIZE_MAX;
+    }
+    return m_spectrums.front()->stamp;
+}
+
+void SpectrogramMatcher::conv(const float* src, size_t num, float* dst){
+    if (num < m_spikeKernel.size()){
+        return;
+    }
+    for(size_t i = 0; i < num-m_spikeKernel.size()+1; i++){
```
Mysticial, line 63: Reminder for myself to profile this step to make sure it's not consuming an unreasonable amount of CPU.

PR #101, opened 2022-03-20
Adding UnownFinder program for PLA

PhobosLegacy:


SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_UnownFinder.cpp
```diff
@@ -0,0 +1,288 @@
...
+    std::atomic<uint64_t>& attempts;
+    std::atomic<uint64_t>& errors;
+    std::atomic<uint64_t>& shinies;
+};
+
+std::unique_ptr<StatsTracker> UnownFinder::make_stats() const{
+    return std::unique_ptr<StatsTracker>(new Stats());
+}
+
+void UnownFinder::enter_map(SingleSwitchProgramEnvironment& env){
```
Mysticial, line 88: Why not just use `goto_camp_from_jubilife()`?

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_UnownFinder.cpp
```diff
@@ -0,0 +1,288 @@
...
+    );
+    if (ret < 0){
+        stats.errors++;
+        throw OperationFailedException(env.console, "Map not detected after 10 x A presses.");
+    }
+    env.console.log("Found map!");
+    env.wait_for(std::chrono::milliseconds(500));
+}
+
+bool UnownFinder::select_map(SingleSwitchProgramEnvironment &env, MapRegion &region){
```
Mysticial, line 111: Same here, this is already handled by `goto_camp_from_jubilife()`.

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_UnownFinder.cpp
```diff
@@ -0,0 +1,288 @@
...
+        pbf_press_dpad(env.console, DPAD_RIGHT, 20, 40);
+        env.console.botbase().wait_for_all_requests();
+    }
+
+    //Enter region
+    mash_A_to_change_region(env, env.console);
+    return true;
+}
+
+bool UnownFinder::pick_mount(SingleSwitchProgramEnvironment &env, MountState &mount_on, MountState &mount_off){
```
Mysticial, line 140: `change_mount()` does this entire function for you.

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_UnownFinder.cpp
```diff
@@ -0,0 +1,288 @@
...
+    MountState target_mount_on = MountState::BRAVIARY_ON;
+    MountState target_mount_off = MountState::BRAVIARY_OFF;
+    if (!pick_mount(env, target_mount_on, target_mount_off)) return false;
+
+    //Start path
+    env.console.log("Beginning Shiny Detection...");
+    {
+        ShinySoundDetector shiny_detector(env.console, SHINY_DETECTED.stop_on_shiny());
+
+        if(SKIP_PATH_SHINY){
```
Mysticial, line 205: If you transpose this function a bit, you can have `run_until()` run while ignoring shinies if you set the flag to false. So even though it's ignoring shinies, it'll still report them and add them to the stats.

So what you could do here is to have two separate `run_until()` sessions. The first is the route to the ruins. That one you can set the stop-on-detected flag according to whether the user wants to ignore shinies. Once you reach the ruins, you can launch a 2nd `run_until()` with shiny stop enabled to run into the ruins.

PhobosLegacy: Hi, I'm changing to follow this approach but I'm having some issues on the `on_shiny_sound` as it expects a `ShinyDetectedActionOption`. 
I'm trying to create a new one but I can't seem to find a way to set the action to `IGNORE`

This is how it looks:

```
        ShinyDetectedActionOption SHINY_DETECTED_ON_ROUTE("0");
        ShinySoundDetector shiny_detector_route(env.console, !SKIP_PATH_SHINY );
        run_until(
            env, env.console,
            [](const BotBaseContext& context){
               ruins_entrance_route(context);
            },
            { &shiny_detector_route }
        );

        if (shiny_detector_route.detected()){
           stats.shinies++;
           on_shiny_sound(env, env.console, SHINY_DETECTED_ON_ROUTE, shiny_detector_route.results());
        }
```
But because the new `SHINY_DETECTED_ON_ROUTE` have no ACTION set, it also stops when hearing a shiny. Only way I could think was to convert the `SKIP_PATH_SHINY` boolean to a `ShinyDetectedActionOption` and let the user set like for the other one.
Any suggestions?

PR #117, opened 2022-04-18
Shiny Tree Hunt Program

PhobosLegacy:


SerialPrograms/Source/PokemonLA/Programs/PokemonLA_RegionNavigation.cpp
```diff
@@ -380,6 +380,94 @@ void goto_camp_from_overworld(
...
+                }
+                for (size_t c = 0; c < location.warp_sub_slot; c++){
+                    pbf_press_dpad(context, DPAD_DOWN, 20, 30);
+                }
+                pbf_mash_button(context, BUTTON_A, 125);
+                break;
+            }
+
+            if (ret < 0){
+                throw OperationFailedException(console, "Unable to fly. Are you under attack?");
```
Mysticial, line 449: Is this intentional? `EscapeFromAttack` is not reliable in that when it returns you may still be under attack. Thus the loop that you probably based this from is a loop so that it can retry if it still can't fly.

SerialPrograms/Source/PokemonLA/Programs/PokemonLA_RegionNavigation.cpp
```diff
@@ -380,6 +380,94 @@ void goto_camp_from_overworld(
...
+            );
+            int ret = wait_until(
+                console, context,
+                std::chrono::seconds(2),
+                {{detector}}
+            );
+            if (ret >= 0){
+                console.log("Flying back to camp...");
+                pbf_wait(context, 50);
+                if (location.warp_slot != 0){
```
Mysticial, line 436: I don't think this does what you think it does. `warp_slot` is the slot when viewed from Jubilife. `warp_subslot` is the slot when viewed from inside the region. So for example, arenas and the Diamond/Pearl settlements are accessible via `warp_subslot`, but not via `warp_slot` from Jubilife.

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_TreeShinyHunting.h
```diff
@@ -0,0 +1,53 @@
...
+ *
+ */
+
+#ifndef PokemonAutomation_PokemonLA_TreeShinyHunting_H
+#define PokemonAutomation_PokemonLA_TreeShinyHunting_H
+
+#include "CommonFramework/Notifications/EventNotificationsTable.h"
+#include "NintendoSwitch/Framework/NintendoSwitch_SingleSwitchProgram.h"
+#include "PokemonLA/Options/PokemonLA_ShinyDetectedAction.h"
+#include "PokemonLA/Inference/PokemonLA_MountDetector.h"
```
Mysticial, line 13: Is this included needed?

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_TreeShinyHunting.cpp
```diff
@@ -0,0 +1,289 @@
...
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include "Common/Cpp/Exceptions.h"
+#include "CommonFramework/Notifications/ProgramNotifications.h"
+#include "CommonFramework/Tools/StatsTracking.h"
+#include "CommonFramework/InferenceInfra/InferenceRoutines.h"
+#include "CommonFramework/Tools/ErrorDumper.h"
```
Mysticial, line 11: How many of these are actually needed?

SerialPrograms/Source/PokemonLA/Programs/PokemonLA_RegionNavigation.cpp
PhobosLegacy: yeah, removed, I think was a left over from when I was playing around with it!

SerialPrograms/Source/PokemonLA/Programs/PokemonLA_RegionNavigation.cpp
PhobosLegacy: huum not sure, I got the idea of warp_slot being for the camps (accessible from Jubilife) and the subslot for the remaining locations. But since I'm using ```TravelLocations::instance()``` to travel, it won't work without that first block as it checks the first integer in the TravelLocation

eg. If I remove 
```
if (location.warp_slot != 0){
                    for (size_t c = 0; c < location.warp_slot; c++){
                        pbf_press_dpad(context, DPAD_DOWN, 20, 30);
                    }
                }
```

and try to go to Fieldlands_Heights (1,0), it will not work, it will only take in consideration the 0 (since the 1 was not used)

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_TreeShinyHunting.h
PhobosLegacy: removed the ones not used! Thanks!

SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_TreeShinyHunting.cpp
PhobosLegacy: removed!

SerialPrograms/Source/PokemonLA/Programs/PokemonLA_RegionNavigation.cpp
Mysticial: Oh yeah you're right. I forgot that the sub-warp-slot is only when you've arrive.

SerialPrograms/CMakeLists.txt.user
```diff
@@ -0,0 +1,322 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Mysticial, line 1: Can you remove this compiler artifact?

Mysticial:
Looks good so far. Can you try rebasing to eliminate all the line-ending diffs?

SerialPrograms/CMakeLists.txt.user
PhobosLegacy: removed!

PhobosLegacy:
I think I managed to rebase, the big reds scared me in the diffs but I believe everything is in order!

PR #129, opened 2022-05-07
use enum for tree actions

pifopi:
Use an enum instead of hardcoded values


PhobosLegacy:
Small suggestion (but not big of a deal) would be to rename from TreePokemon to something more generic (maybe LeapPokemon?) since it also include the ones from Ore. But again, very not important :3 
Thanks for this!

pifopi:
> Small suggestion (but not big of a deal) would be to rename from TreePokemon to something more generic (maybe LeapPokemon?) since it also include the ones from Ore. But again, very not important :3 Thanks for this!

Indeed, that's much better ! Should be fine now, i also renamed the files and the method

PR #131, opened 2022-05-08
add ingo move grinder

pifopi:
I added a draft of ingo move grinder to share my code but it's not ready to share to public !

### DIFF
--- a/SerialPrograms/Source/PokemonLA/Options/PokemonLA_BattlePokemonActionTable.cpp
+++ b/SerialPrograms/Source/PokemonLA/Options/PokemonLA_BattlePokemonActionTable.cpp
@@ -9,9 +9,11 @@
 #include "Common/Qt/QtJsonTools.h"
 #include "CommonFramework/Globals.h"
 #include "CommonFramework/Options/EditableTableOption-EnumTableCell.h"
+#include "CommonFramework/Options/EnumDropdownWidget.h"
 #include "Pokemon/Options/Pokemon_IVCheckerWidget.h"
 #include "PokemonLA_BattlePokemonActionTable.h"
 
+#include <QLabel>
 #include <iostream>
 using std::cout;
 using std::endl;
@@ -103,7 +105,7 @@ std::unique_ptr<EditableTableRow> BattlePokemonActionTableFactory::make_row() co
 
 std::vector<std::unique_ptr<EditableTableRow>> BattlePokemonActionTable::make_defaults() const{
     std::vector<std::unique_ptr<EditableTableRow>> ret;
-    ret.emplace_back(std::unique_ptr<BattlePokemonActionRow>(new BattlePokemonActionRow()));
+    ret.emplace_back(std::make_unique<BattlePokemonActionRow>());
     return ret;
 }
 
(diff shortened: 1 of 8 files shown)


Common/Qt/Options/EditableTable/EditableTableBaseOption.h
```diff
@@ -56,6 +56,7 @@ class EditableTableBaseOption{
 
     size_t size() const{ return m_current.size(); }
     const EditableTableRow& operator[](size_t index) const;
+    EditableTableRow& operator[](size_t index);
```
pifopi, line 59: Can I do this ? Else, I can let the config stay constant and have instead the amount of moves I did count up, until it reach the config

SerialPrograms/Source/CommonFramework/Options/EditableTableOption.cpp
```diff
@@ -121,4 +121,10 @@ QWidget* make_double_table_cell(QWidget& parent, double& value, double min, doub
 template QWidget* make_integer_table_cell(QWidget& parent, uint16_t& value);
 template QWidget* make_integer_table_cell(QWidget& parent, int16_t& value);
 
+QWidget* make_conststring_table_cell(QWidget& parent, const QString& string) {
```
pifopi, line 124: Not really an editable but is there a way to use the table and input constant text in it ? (I could use QLabel too)

SerialPrograms/Source/PokemonLA/Options/PokemonLA_BattlePokemonActionTable.cpp
```diff
@@ -251,6 +251,109 @@ MoveStyle OneMoveBattlePokemonActionTable::get_style(size_t pokemon){
...
+std::vector<std::unique_ptr<EditableTableRow>> MoveGrinderActionTable::make_defaults() const {
+    std::vector<std::unique_ptr<EditableTableRow>> ret;
+    for (int i = 0; i < 5; ++i)
+    {
+        for (int j = 0; j < 4; ++j)
+        {
+            ret.emplace_back(std::make_unique<MoveGrinderActionRow>(STRING_POKEMON + " " + QString::number(i), "Move " + QString::number(j)));
+        }
+    }
+    return ret;
```
pifopi, line 354: Is there a way to "freeze" the table, meaning disable the add/remove, etc ? If not, maybe I should create a new structure for this and the issue above

SerialPrograms/Source/PokemonLA/Programs/Farming/PokemonLA_IngoMoveGrinder.cpp
```diff
@@ -0,0 +1,394 @@
+/*  Ingo Move Grinder
```
pifopi, line 1: Lot of duplication in this file, I'll try to reduce that a bit

SerialPrograms/Source/PokemonLA/Options/PokemonLA_BattlePokemonActionTable.cpp
pifopi: Other issue is fixed and hopefully the current display will make it clear that the user isn't supposed to add/remove lines but well

PR #132, opened 2022-05-08
berryTree

Gin890:
Under-develop berry tree detector

### DIFF
--- a/Common/Cpp/Color.h
+++ b/Common/Cpp/Color.h
@@ -14,10 +14,21 @@
 namespace PokemonAutomation{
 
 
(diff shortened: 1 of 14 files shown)


Common/Cpp/Color.h
```diff
@@ -29,6 +40,11 @@ class Color{
         return m_argb < color.m_argb;
     }
 
+    uint8_t a() const { return (m_argb & 0xff000000) >> 24; }
```
Mysticial, line 43: Optimization: Do `(uint8_t)(m_argb >> 24);` instead.

PR #143, opened 2022-05-22
burmy path improvements part 1

PhobosLegacy:


SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_BurmyFinder.cpp
```diff
@@ -268,6 +268,182 @@ void BurmyFinder::go_to_height_camp(SingleSwitchProgramEnvironment& env, BotBase
...
+                pbf_press_button(context, BUTTON_Y, (2.1 * TICKS_PER_SECOND), 0);
+                // Get off Braviary
+                pbf_press_button(context, BUTTON_PLUS, 20, (1 * TICKS_PER_SECOND));
+                // Move camera down
+                pbf_move_right_joystick(context, 128, 255, (0.1 * TICKS_PER_SECOND), (0.3 * TICKS_PER_SECOND));
+
+                // Now we should be on ground
+                disable_shiny_sound(context);
+                check_tree_no_stop(env, context);
+                context.wait_for_all_requests();
```
Gin890, line 328: Battle menu detector might find a battle during the wait here. But `last_checked_tree` is still 1. What about moving the assignment of `last_checked_tree` before `check_tree_no_stop()`?
Same to other places where `last_checked_tree` is changed.

PhobosLegacy: good point, changed!

PR #148, opened 2022-05-26
More Trees

PhobosLegacy:


SerialPrograms/Source/PokemonLA/Programs/ShinyHunting/PokemonLA_BurmyFinder.cpp
```diff
@@ -134,7 +134,7 @@ std::unique_ptr<StatsTracker> BurmyFinder::make_stats() const{
 
 
 struct BurmyFinder::TreeCounter{
-    std::array<uint64_t, 10> tree = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+    std::array<uint64_t, 16> tree = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
```
pifopi, line 137: std::array<uint64_t, 16> tree = {}
or
std::array<uint64_t, 16> tree = { 0 }
is enough to init them all to 0 ;)

PhobosLegacy: Removed all of them :) 
Thanks!

PR #156, opened 2022-06-14
Bdsp gender filter

PhobosLegacy:


SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxGenderDetector.cpp
```diff
@@ -0,0 +1,102 @@
...
+    //logger.log("FEMALE RMSD:" + std::to_string(rmsd));
+    return rmsd <= 140;
+}
+
+
+EggHatchGenderFilter BoxGenderDetector::identify_gender(LoggerQt& logger, VideoOverlay& overlay,const QImage& frame)
+{
+    InferenceBoxScope gender_box(overlay, 0.733, 0.022, 0.204, 0.049, COLOR_BLUE);
+
+    QImage name_and_gender = extract_box_reference(frame, gender_box).to_qimage();
```
Mysticial, line 71: Since you're not changing the image, you can save the reference instead to avoid copying it.

`ConstImageRef image = extract_box_reference(frame, gender_box);`

SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxGenderDetector.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    PackedBinaryMatrix2 matrix = compress_rgb32_to_binary_range(name_and_gender, 0, 130, 0, 130, 0, 255);
+    //logger.log(matrix.dump());
+    std::vector<WaterfillObject> objects = find_objects_inplace(matrix, 200);
```
Mysticial, line 75: We don't really need to do it here since the image is small. But in larger images, you may get a large number of objects. So there's an alternate API where you can iterate them as they are found as opposed to returning everything into a single vector and then operating on them later.

Mysticial: Also. Is 200 too large? The icon is quite small.

SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxGenderDetector.cpp
PhobosLegacy: changed!

SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxGenderDetector.cpp
PhobosLegacy: reduced to 50 !

PR #158, opened 2022-06-18
Adding Program to test movement

PhobosLegacy:


SerialPrograms/Source/CommonFramework/Options/EditableTableOption.cpp
```diff
@@ -90,6 +90,34 @@ template<typename T> QWidget* make_integer_table_cell(QWidget& parent, T& value)
     return box;
 }
 
+
+QWidget* make_limited_integer_table_cell(QWidget& parent, uint8_t& value, uint8_t min, uint8_t max){
```
Mysticial, line 94: Do we need to limit this to only 8-bit integers?

PhobosLegacy: changed to int :)

SerialPrograms/Source/CommonFramework/Options/EditableTableOption.cpp
```diff
@@ -90,6 +90,34 @@ template<typename T> QWidget* make_integer_table_cell(QWidget& parent, T& value)
     return box;
 }
 
+
+QWidget* make_limited_integer_table_cell(QWidget& parent, int& value, int min, int max){
```
Mysticial, line 94: Can we actually change this to either `int32_t` or `int64_t`. The size of `int` is not defined for C++.

PhobosLegacy: done!

PR #159, opened 2022-06-28
Adding Ramanas Island Combee Finder program + Fix on TestPathMaker

PhobosLegacy:


SerialPrograms/Source/PokemonLA/Programs/General/PokemonLA_RamanasIslandCombee.cpp
```diff
@@ -0,0 +1,315 @@
...
+            pbf_move_left_joystick(context, 241, 0, 100, 20);
+            pbf_press_button(context, BUTTON_B, 2390, 0);
+            pbf_press_button(context, BUTTON_Y, 380, 0);
+            pbf_move_right_joystick(context, 127, 255, 90, 20);
+            check_tree_no_stop(env, context);
+
+            env.console.log("Checking Tree 2");
+            pbf_press_button(context, BUTTON_PLUS, 20, 200);
+            pbf_move_left_joystick(context, 242, 0, 100, 20);
+            enable_shiny_sound(context);
```
Mysticial, line 179: Since the button presses are async, you can't reliably schedule the enable-shiny-sound here. You'll need to precede it with a wait-for-all. However, once you do that, the shiny sound won't turn back on until after you've already started moving again - thus creating a window to missing an enroute shiny. So you'll actually want to move it up one line.

```
pbf_press_button(context, BUTTON_PLUS, 20, 200);
context.wait_for_all_requests();
enable_shiny_sound(context);
pbf_move_left_joystick(context, 242, 0, 100, 20);
```

PhobosLegacy: got it! 
done!

PR #163, opened 2022-07-24
Migrate most of MapSpriteReader to new image API.

Mysticial:
I'm a bit scared to touch this file. So here's a PR.

I didn't touch the HSV stuff yet since I think that might require a bit of redesign.


Gin890 (approved):
LGTM

PR #170, opened 2022-08-18
Add SwSh BerryFarmer which uses sound as feedback

fyex:
Unlike the original Berry Farmer, this program requires you to have a Pokemon in the first party slot which can escape from all tree encounters and OHKO them.
This is why I decided to make it a seperate program. Though, if it is preferred it wouldn't be hard to incorporate it into the original program as an option.

### DIFF
--- a/SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
+++ b/SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
@@ -4,18 +4,10 @@
  *
  */
 
-#include <QString>
+
 #include "CommonFramework/Inference/SpectrogramMatcher.h"
 #include "CommonFramework/Inference/AudioTemplateCache.h"
-#include "CommonFramework/Tools/ConsoleHandle.h"
-//#include "PokemonSwSh/PokemonSwSh_Settings.h"
 #include "PokemonSwSh_BerryTreeRustlingSoundDetector.h"
-
-#include <sstream>
-#include <cfloat>
-#include <iostream>
-using std::cout;
-using std::endl;
 
 namespace PokemonAutomation{
 namespace NintendoSwitch{
--- a/SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
+++ b/SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
@@ -1,0 +1,48 @@
+﻿/*  Berry Tree Rustling Sound Detector
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include <QString>
+#include "CommonFramework/Inference/SpectrogramMatcher.h"
+#include "CommonFramework/Inference/AudioTemplateCache.h"
+#include "CommonFramework/Tools/ConsoleHandle.h"
+//#include "PokemonSwSh/PokemonSwSh_Settings.h"
+#include "PokemonSwSh_BerryTreeRustlingSoundDetector.h"
+
+#include <sstream>
+#include <cfloat>
+#include <iostream>
+using std::cout;
+using std::endl;
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonSwSh{
+
+
+BerryTreeRustlingSoundDetector::BerryTreeRustlingSoundDetector(Logger& logger, ConsoleHandle& console, OnShinyCallback on_shiny_callback)
+    // Use a red as the detection color because the Levelball is made using red apricorns.
+    : AudioPerSpectrumDetectorBase(logger, "BerryTreeRustlingSoundDetector", "Berry tree rustling sound", COLOR_RED, console, on_shiny_callback)
+{}
+
+
+float BerryTreeRustlingSoundDetector::get_score_threshold() const{
+    return 0.68f;
+}
+
+std::unique_ptr<SpectrogramMatcher> BerryTreeRustlingSoundDetector::build_spectrogram_matcher(size_t sampleRate){
+    return std::make_unique<SpectrogramMatcher>(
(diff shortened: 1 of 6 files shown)


SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
```diff
@@ -0,0 +1,48 @@
+﻿/*  Berry Tree Rustling Sound Detector
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include <QString>
```
pifopi, line 7: minor but I think some header could be cleaned up !

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
```diff
@@ -0,0 +1,275 @@
...
+    PA_ADD_OPTION(SECONDARY_ATTEMPT_MASH_TIME);
+}
+
+
+BerryFarmer2::Rustling BerryFarmer2::check_rustling(SingleSwitchProgramEnvironment& env, BotBaseContext& context) {
+    // wait some time in order to not detect rustling from previous fetch attempt
+    pbf_wait(context, 80);
+    context.wait_for_all_requests();
+
+    std::chrono::time_point initial_rustling_time = std::chrono::system_clock::now();
```
pifopi, line 101: could it be declared line 128 directly ?

pifopi (approved):
Looks good to me ! But I'll let Kuro and / or Gin review this because they're way more knowledgeable on it !

SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
Mysticial: Yeah, we're in the process of removing Qt from anything that isn't UI code. So you shouldn't need `QString` unless you need some of its library functions which are available elsewhere..

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
Mysticial: `Common/Cpp/Time.h` has `WallClock` and `current_time()` if you want to avoid writing the full chrono stuff.

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    }
+    else if (ret == 1) {
+        env.console.log("BerryFarmer: Battle Menu detected!");
+        result = Rustling::Battle;
+    } 
+    else if (ret == 2) {
+        env.console.log("BerryFarmer: Battle Start detected.");
+        // waiting here so we can be sure the battle is fully underway so escaping is possible
+        pbf_wait(context, START_BATTLE_DELAY);
```
Mysticial, line 157: Is this case rare? Blind waiting for the battle to fully start seems suboptimal? There's a `StandardBattleMenuWatcher` that will wait for it with feedback.

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+        if (current_rustling == Rustling::No) {
+            no_rustling++;
+        }
+    }
+    /* Fast rustling, in Battle or too many times No rustling */
+    if (no_rustling >= 3) {
+        return attempts;
+    }
+    if (current_rustling == Rustling::Fast) {
```
Mysticial, line 190: If I'm reading this right, you're shaking the tree even though it's fast rustling. (thus risking a battle) Is this somehow a calculated optimal algorithm?

SerialPrograms/Source/PokemonSwSh/Inference/Sounds/PokemonSwSh_BerryTreeRustlingSoundDetector.cpp
fyex: fixed

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
fyex: moved declaration to line 128 and changed to WallClock for readability.

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
fyex: A Battle should be quite rare. Changed to `StandardBattleMenuWatcher` anyway since it was already used elswehere.

SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/PokemonSwSh_DateSpam-BerryFarmer2.cpp
fyex: Yes, if there is at least one slow rustling first there will be an additional fetch attempt at the end. While I don't have exact numbers it is quite rare for a battle to happen after the first fast rustling (unless it's right after the very first fetch attempt).
This way should be slightly more efficient if battles are less than one in 20ish extra attempts. 
During my tests battles were less than one in 40.

PR #171, opened 2022-08-21
save and load profiles for console settings

pifopi:


SerialPrograms/Source/NintendoSwitch/Framework/UI/NintendoSwitch_SwitchSystemWidget.cpp
```diff
@@ -98,6 +100,35 @@ SwitchSystemWidget::SwitchSystemWidget(
...
+            }
+            m_session.option().load_json(load_json_file(path));
+            // When playing with profile, we are necessarly in a stopped state
+            update_ui(ProgramState::STOPPED);
+            m_session.audio().reset();
+            m_session.camera_session().set_source(CameraInfo(m_session.option().m_camera.info));
+            m_session.serial_session().botbase().try_reset();
+            m_session.video().reset();
+            m_serial_widget->reset();
+            m_camera_widget->update_ui();
```
Mysticial, line 118: I see what you're doing here. You made `option()` public so you can directly deserialize the state onto it. But that bypasses the entire session class which ensure thread-safety as well as automatically UI. You're not observing the thread-safety issues because you're lucky and the UI doesn't update because you bypassed it.

The cleanest way to properly do what you want here to construct a new `SwitchSystemOption` and deserialize the JSON to that. Then update the session. So something like this:

```
//  Will need to add a method to `SwitchSystemSession` to fetch `min_pabotbase` from the underlying option. This is thread-safe since the field is const.
SwitchSystemOption option(m_session.min_pabotbase(), m_session.allow_commands_while_running());

//  Deserialize into this local option instance.
option.load_json(load_json_file(path));

m_session.serial_session().botbase().reset(option.m_serial.port());
m_session.camera_session().set_source(option.m_camera.info);
m_session.camera_session().set_resolution(option.m_camera.current_resolution);
if (!option.m_audio.input_file().empty()){
    m_session.audio_session().set_audio_input(option.m_audio.input_file());
}else{
    m_session.audio_session().set_audio_input(option.m_audio.input_device());
}
m_session.audio_session().set_audio_format(option.m_audio.input_format());
m_session.audio_session().set_audio_output(option.m_audio.output_device());
m_session.audio_session().set_volume(option.m_audio.volume());
m_session.audio_session().set_display(option.m_audio.display_type());
```

Basically all interaction is through the session classes. They will handle the thread-safety and updating the UI.

pifopi: thanks ! I did this and some small changes (there was an infinite loop with audio selector and the combobox for serial wasn't updating when loading from profile)

SerialPrograms/Source/NintendoSwitch/Framework/NintendoSwitch_SwitchSystemSession.h
```diff
@@ -60,6 +61,7 @@ class SwitchSystemSession final : public TrackableConsole{
     CameraSession& camera_session(){ return *m_camera; }
     AudioSession& audio_session(){ return m_audio; }
     VideoOverlaySession& overlay_session(){ return m_overlay; }
+    const SwitchSystemOption& option() const{ return m_option;  }
```
pifopi, line 64: forgot to remove this :(

PR #174, opened 2022-09-03
Various fixes

pifopi:


Common/Qt/Options/TextEditWidget.cpp
```diff
@@ -40,7 +40,7 @@ class TextEditWidget::Box : public QTextEdit{
 #if 0
         connect(
             this, &QTextEdit::textChanged,
-            [=]{
+            []{
```
Gin890, line 43: Need this too.
It's in a if 0 so not compiled. But better to be complete.

PR #191, opened 2022-09-21
add ball reading capability to home sorter

pifopi:


SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -117,13 +128,34 @@ struct Cursor{
...
-    (*arr)[pos1] = (*arr)[pos2];
-    (*arr)[pos2] = temp;
+bool operator<(const Pokemon& lhs, const Pokemon& rhs) {
+    // For example, we could image that people want to sort first by ball type, then by dex number
+    if (lhs.ball_slug < rhs.ball_slug){
+        return true;
+    }
+    if (lhs.ball_slug > rhs.ball_slug){
+        return false;
+    }
```
pifopi, line 149: I won't commit this when merging. that way, even if it read balls, it won't change the default behaviour

PR #206, opened 2022-11-11
Home sorter nav improvements, sort options table and json output

jtszalay:
This PR adds the following:

1.  Row and Column nav now use shortest path with wrap around when it makes sense to do so.
2. A table for options related to sort order (need to handle the "reverse" checkbox)
3. JSON file output of current state and sorted state.
4. "Dry run" option for just collecting current state (and outputting file showing sorted state if it were to be applied with current settings.)

### DIFF
--- a/SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
+++ b/SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
@@ -25,6 +25,7 @@ const std::set<std::string> TOKENS{
     "6643d9fe87b3e54dc75dfac8ac22f0cc8bd17f6a8a786debf5fc4c517ee65469",
     "8e48e38e49bffc8462ada9d2d9d850d5b3b5c9529d20978c09bc548bc9a614a4",
     "7694adee4419d62c6a923c4efc9e7b41def7b96bb84ea882701b0bf2e8c13bee",
+    "e8d168bc482e96553ea9f9ecaea5a817474dbccc2a6a228a6bde67f2b2aa2889",
 };
 
 
--- a/SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
+++ b/SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
@@ -113,8 +113,8 @@ BoxSorting::BoxSorting()
           false,
           LockWhileRunning::LOCKED,
           "<b>Output File:</b><br>JSON file for output of storage boxes.",
-          "box_order.json",
-          "box_order.json"
+          "box_order",
+          "box_order"
       )
     , DRY_RUN(
           "<b>Dry Run:</b><br>Catalogue and make sort plan without executing. (Will output to OUTPUT_FILE and OUTPUT_FILE.sortplan)",
@@ -393,7 +393,7 @@ void output_boxes_data_json(const std::vector<std::optional<Pokemon>>& boxes_dat
         }
         pokemon_data.push_back(std::move(pokemon));
     }
-    pokemon_data.dump(json_path);
+    pokemon_data.dump(json_path + ".json");
 }
 
 void do_sort(
@@ -707,7 +707,7 @@ void BoxSorting::program(SingleSwitchProgramEnvironment& env, BotBaseContext& co
 
     env.console.log("Sorted boxes data :");
     print_boxes_data(boxes_sorted, env);
-    const std::string sorted_path = json_path + ".sorted";
+    const std::string sorted_path = json_path + "-sorted";
     output_boxes_data_json(boxes_sorted, sorted_path);
 
     if (!DRY_RUN) {
--- a/SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
+++ b/SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
@@ -397,6 +397,62 @@ void output_boxes_data_json(const std::vector<std::optional<Pokemon>>& boxes_dat
     pokemon_data.dump(json_path);
 }
 
+void do_sort(
+        SingleSwitchProgramEnvironment& env,
+        BotBaseContext& context,
+        std::vector<std::optional<Pokemon>> boxes_data,
+        std::vector<std::optional<Pokemon>> boxes_sorted,
+        BoxSorting_Descriptor::Stats& stats,
+        Cursor& cur_cursor,
+        uint16_t GAME_DELAY
+        ) {
+    std::ostringstream ss;
+    // this need to be separated into functions when I will redo the whole thing but I just wanted it to work
+
+    // going thru the sorted list one by one and for each one go through the current pokemon layout to find the closest possible match to fill the slot
+    for (size_t poke_nb_s = 0; poke_nb_s < boxes_sorted.size(); poke_nb_s++){
+        if (boxes_sorted[poke_nb_s] == std::nullopt){ // we've hit the end of the sorted list.
+            break;
+        }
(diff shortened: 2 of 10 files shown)


SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
```diff
@@ -25,6 +25,7 @@ const std::set<std::string> TOKENS{
     "6643d9fe87b3e54dc75dfac8ac22f0cc8bd17f6a8a786debf5fc4c517ee65469",
     "8e48e38e49bffc8462ada9d2d9d850d5b3b5c9529d20978c09bc548bc9a614a4",
     "7694adee4419d62c6a923c4efc9e7b41def7b96bb84ea882701b0bf2e8c13bee",
+    "e8d168bc482e96553ea9f9ecaea5a817474dbccc2a6a228a6bde67f2b2aa2889",
```
pifopi, line 28: You can modify directly the code in SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp / .h to directly change DEVELOPER_MODE bool, much easier than adding your own token ;)

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -168,30 +196,56 @@ bool operator<(const std::optional<Pokemon>& lhs, const std::optional<Pokemon>&
...
-    // }
-    return lhs->dex_number < rhs->dex_number;
+
+    for (const BoxSortingSelection preference : *lhs->preferences){
+        switch(preference.sort_type) {
+            // NOTE edit when adding new struct members
+            // TODO TESTING and account for preference.reverse
+            case BoxSortingSortType::NationalDexNo:
+                return lhs->national_dex_number < rhs->national_dex_number;
+                break;
```
pifopi, line 206: if lhs->national_dex_number == rhs->national_dex_number, there is a logical bug because we'd like to break out of the switch instead of returning false directly (there may be more checks later on)

```suggestion
            case BoxSortingSortType::NationalDexNo:
                if (lhs->national_dex_number != rhs->national_dex_number){
                    return lhs->national_dex_number < rhs->national_dex_number;
                }
                break;
```

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -292,13 +372,50 @@ void print_boxes_data(const std::vector<std::optional<Pokemon>>& boxes_data, Sin
...
+            // NOTE edit when adding new struct members
+            pokemon["national_dex_number"] =  boxes_data[poke_nb]->national_dex_number;
+            pokemon["shiny"] =  boxes_data[poke_nb]->shiny;
+            pokemon["gmax"] =  boxes_data[poke_nb]->gmax;
+            pokemon["ball_slug"] =  boxes_data[poke_nb]->ball_slug;
+            pokemon["gender"] = gender_to_string(boxes_data[poke_nb]->gender);
+        }
+        pokemon_data.push_back(std::move(pokemon));
+    }
+    if (sorted) {
```
pifopi, line 394: I would move the fact to add ".sorted" to the calling site but maybe it's only my preference, feel free to do what you prefer !

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -292,13 +372,50 @@ void print_boxes_data(const std::vector<std::optional<Pokemon>>& boxes_data, Sin
...
+        pokemon_data.dump(json_path);
+    }
+
+}
+
 void BoxSorting::program(SingleSwitchProgramEnvironment& env, BotBaseContext& context){
+
+    std::vector<BoxSortingSelection> sort_preferences = SORT_TABLE.preferences();
+    if (sort_preferences.empty()) {
+        throw UserSetupError(env.console, "At least sorting selection needs to be made!");
```
pifopi, line 407: one is missing ?

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -147,18 +170,23 @@ size_t get_index(size_t box, size_t row, size_t column){
 
 
 struct Pokemon{
-    // When adding any new member here, do not forget to modify the three operators below
-    uint16_t dex_number = 0;
+    std::vector<BoxSortingSelection>* preferences;
```
pifopi, line 173: could you use a const reference instead ? It seems to never be null / changed

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -170,14 +168,15 @@ size_t get_index(size_t box, size_t row, size_t column){
 
 
 struct Pokemon{
-    std::vector<BoxSortingSelection>* preferences;
+    const std::vector<BoxSortingSelection>* preferences;
```
pifopi, line 171: I meant a real reference (everything that could be a reference instead of a ptr should be imo, that way you're sure it is never null when using it) :
const std::vector<BoxSortingSelection>&
but that is fine nonetheless

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -9,10 +9,8 @@ break into smaller functions
 read pokemon name and store the slug (easier to detect missread than reading a number)
 Optimise the swapping algo
 Add enum for ball ? Also, BDSP is reading from swsh data. Worth refactoring ?
-Find a way to let the user customize the sorting algo based on their preference
 
 ideas for more checks :
-gender
 ability
 nature
 type
```
pifopi, line 16: remove id ot from the list

pifopi (approved):

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
jtszalay: Oh! I can do that!

jtszalay: PokemonHome_BoxSorting.cpp:445: error: C2664: 'void std::swap(std::thread &,std::thread &) noexcept': cannot convert argument 1 from '_Ty' to 'std::thread &'

Seems swap is unhappy when I do that.

pifopi: oh true, my bad !

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -292,13 +373,106 @@ void print_boxes_data(const std::vector<std::optional<Pokemon>>& boxes_data, Sin
...
+            pokemon["national_dex_number"] =  boxes_data[poke_nb]->national_dex_number;
+            pokemon["shiny"] =  boxes_data[poke_nb]->shiny;
+            pokemon["gmax"] =  boxes_data[poke_nb]->gmax;
+            pokemon["ball_slug"] =  boxes_data[poke_nb]->ball_slug;
+            pokemon["gender"] = gender_to_string(boxes_data[poke_nb]->gender);
+            pokemon["ot_id"] = boxes_data[poke_nb]->ot_id;
+        }
+        pokemon_data.push_back(std::move(pokemon));
+    }
+    pokemon_data.dump(json_path);
```
pifopi, line 396: I would add the .json here. That way, we keep the right extension for the sorted version. The json is awesome

PR #207, opened 2022-11-16
use only one loop per fetch for swsh autonomous eggs

pifopi:


pifopi:
to be merged once the freeze is lifted !

PR #231, opened 2022-12-08
Add LP farming mode to Tera Self Farmer

ScottehMax:
This will currently always skip the raid when this setting is on.


SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraSelfFarmer.cpp
```diff
@@ -277,6 +278,14 @@ void TeraSelfFarmer::program(SingleSwitchProgramEnvironment& env, BotBaseContext
 //            env.log("Detected a " + TERA_TYPE_NAMES[(size_t)type] + " " + std::to_string(stars) + " star raid.", COLOR_PURPLE);
         }
 
+        if (MODE == Mode::FARM_LP_ONLY) {
```
pifopi, line 281: I'd say to do this : 

bool skip = false;
if (MODE == Mode::FARM_LP_ONLY){
    skip = true;
}
if (stars < MIN_STARS || stars > MAX_STARS){
    skip = true;
}

if (skip(){
    skip
}

pifopi:
as of now, this is the PR where I red the silhouette but there is no filtering yet, it could totally be added though !

pifopi (approved):

SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraSelfFarmer.cpp
ScottehMax: Done.

PR #240, opened 2022-12-16
gimmi chest farm

kichithewolf:


SerialPrograms/CMakeLists.txt
```diff
@@ -1253,6 +1253,8 @@ file(GLOB MAIN_SOURCES
     Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.h
     Source/PokemonSV/Programs/General/PokemonSV_AutonomousBallThrower.cpp
     Source/PokemonSV/Programs/General/PokemonSV_AutonomousBallThrower.h
+	Source/PokemonSV/Programs/General/PokemonSV_GimmighoulChestFarm.cpp
```
pifopi, line 1256: beware, some tabs are being added here and in other locations

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_GimmighoulChestFarm.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+            //Close the game
+			save_game_from_overworld(env.console, context);
+            pbf_press_button(context, BUTTON_HOME, 20, GameSettings::instance().GAME_TO_HOME_DELAY);
+            context.wait_for_all_requests();
+            close_game(context);
+			
+            //Date skip - in-game day cycle is 72 mins, so 2 hours is fastest way
+			//This isn't perfect because 12 hour format but it works
+            home_to_date_time(context, true, false);
```
pifopi, line 136: hmm, I think this will depends on your clock format. For which format is it ? Maybe Kuro or Gin know a function to handle every clock format ?

pifopi (approved):
beside those two remarsk, all good to me. Thanks

SerialPrograms/CMakeLists.txt
kichithewolf: tabs and spaces fixed with latest commit

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_GimmighoulChestFarm.cpp
kichithewolf: experimented with changing the region/clock format on my switch, it works for both 12 and 24 hour clock layouts (if anything the 24 hour format runs more efficiently). if there's another layout other than those two I'll have to make changes.

12 hour format appears in the settings as month/day/year hour:minute am/pm okbutton.
24 hour format appears in the settings as day/month/year hour:minute okbutton.

hour is in the same position for both formats so moving it forward is fine in both formats. to save the change it presses A 4 times to navigate past minute and am/pm and to press the confirm button. 24 hour format doesn't have am/pm so it ends up pressing A an extra time after it saves the change, which just reopens the date/time menu. the home button press afterwards closes everything and takes it back to menu in both formats.

PR #255, opened 2023-01-12
Release box after hatching in EggAutonomous

chicknbeer:
I copied MassRelease::release_box to BoxRoutines and used it in EggAutonomous

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
@@ -324,7 +324,8 @@ void EggAutonomous::hatch_eggs_full_routine(SingleSwitchProgramEnvironment& env,
     // The loop to hatch batches of eggs.
     // Each batch consists of at most five eggs.
     // There are at most six batches of eggs in a box.
-    uint8_t next_egg_column = 0; // next egg column in box
+    uint8_t next_egg_column = 1; // next egg column in box
+    // starting at 1 because picnic_party_to_hatch_party already loaded and hatched the first batch.
     while(true){
         save_game_if_needed();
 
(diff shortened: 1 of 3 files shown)


SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
```diff
@@ -341,16 +342,20 @@ void EggAutonomous::hatch_eggs_full_routine(SingleSwitchProgramEnvironment& env,
...
         } // end for each hatched pokemon in party
 
+        // move party back into the box
+        env.log("Unload party");
+        env.console.overlay().add_log("Unload party", COLOR_WHITE);
+        unload_one_column_from_party(env.program_info(), env.console, context, next_egg_column-1, HAS_CLONE_RIDE_POKEMON);
+
         // Get the next egg column
         SomethingInBoxSlotDetector sth_in_box_detector(COLOR_RED);
-        for (; next_egg_column < 6; next_egg_column++){
```
chicknbeer, line 346: next_egg_column is incremented on line 417 at the end egg batch loop

SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
```diff
@@ -324,7 +324,8 @@ void EggAutonomous::hatch_eggs_full_routine(SingleSwitchProgramEnvironment& env,
     // The loop to hatch batches of eggs.
     // Each batch consists of at most five eggs.
     // There are at most six batches of eggs in a box.
-    uint8_t next_egg_column = 0; // next egg column in box
+    uint8_t next_egg_column = 1; // next egg column in box
+    // starting at 1 because picnic_party_to_hatch_party already loaded and hatched the first batch.
```
chicknbeer, line 328: A mistake here. picnic_party_to_hatch_party only load the first column into the party.
The first column is not hatched yet. It is hatched during the first iteration at line 336 hatch_eggs_at_zero_gate

SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
```diff
@@ -478,8 +493,6 @@ void EggAutonomous::process_one_baby(SingleSwitchProgramEnvironment& env, BotBas
 
         case EggHatchAction::Release:
```
pifopi, line 494: On phone so hard to check but is this enum value used anywhere else?

pifopi:
Why is it faster to do so ? There is one more step (moving pokemon to the boxes)

SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
chicknbeer: I don't think so. It is the default action set on line 436.

chicknbeer:
> Why is it faster to do so ? There is one more step (moving pokemon to the boxes)

It is faster when you expect to release more often than keeping because the program can restart the game before releasing any pokemon at all. For example breeding only for shiny with the shiny charm, you can expect to one in 512 which means 17 boxes of release would be skipped by skipping straight to restarting the game.

PR #264, opened 2023-01-21
stats reset table instead of egg hatch table

kichithewolf:


SerialPrograms/Source/Pokemon/Options/Pokemon_StatsResetFilter.cpp
```diff
@@ -0,0 +1,144 @@
...
+//#include <iostream>
+//using std::cout;
+//using std::endl;
+
+namespace PokemonAutomation{
+namespace Pokemon{
+
+const EnumDatabase<EggHatchAction>& StatsResetAction_Database() {
+    static const EnumDatabase<EggHatchAction> database({
+        {EggHatchAction::StopProgram,   "stop",     "Stop Program"},
```
Mysticial, line 21: What's the difference between "keep" and "stop" for stats reset?

(I know you copy-pasted this from the egg filters, but the idea is to change places like this.)

SerialPrograms/Source/Pokemon/Options/Pokemon_StatsResetFilter.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {EggHatchAction::Keep,          "keep",     "Keep"},
+        });
+    return database;
+}
+const EnumDatabase<EggHatchShinyFilter>& StatsResetShinyFilter_Database() {
```
Mysticial, line 26: The treasures are all shiny-locked. I don't think we need this?

SerialPrograms/Source/Pokemon/Options/Pokemon_StatsResetFilter.cpp
```diff
@@ -0,0 +1,144 @@
...
+    static const EnumDatabase<EggHatchShinyFilter> database({
+        {EggHatchShinyFilter::Anything, "anything",     "Anything"},
+        {EggHatchShinyFilter::NotShiny, "not-shiny",    "Not Shiny"},
+        {EggHatchShinyFilter::Shiny,    "shiny",        "Shiny"},
+        });
+    return database;
+}
+const EnumDatabase<EggHatchGenderFilter>& StatsResetGenderFilter_Database() {
```
Mysticial, line 34: Same here. Treasures are all genderless. So this doesn't apply either.

SerialPrograms/Source/Pokemon/Options/Pokemon_StatsResetFilter.cpp
```diff
@@ -0,0 +1,144 @@
...
+    , iv_hp(IVCheckerFilter::Anything)
+    , iv_atk(IVCheckerFilter::Anything)
+    , iv_def(IVCheckerFilter::Anything)
+    , iv_spatk(IVCheckerFilter::Anything)
+    , iv_spdef(IVCheckerFilter::Anything)
+    , iv_speed(IVCheckerFilter::Anything)
+{
+    //PA_ADD_OPTION(action);
+    //PA_ADD_OPTION(shiny);
+    //PA_ADD_OPTION(gender);
```
Mysticial, line 56: Oh I see what you did. Let's just clean up all the extra stuff above.

SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.h
```diff
@@ -103,6 +104,13 @@ bool check_baby_info(
     Pokemon::EggHatchAction& action
 );
 
+//This is a similar to check_baby_info, except for Stats Reset
+bool check_stats_reset_info(
```
Mysticial, line 108: Since this is separate from eggs, let's not put it in the same file as the eggs.

PR #265, opened 2023-01-25
Nature filters

kichithewolf:
Nature filter for SwSh, BDSP, and SV egg auto.

Most of the filter code is the IV checker, but with natures.
BDSP and SV read the text, PR in the Packages repo for the json file has also been opened.
SwSh gets the color of the stat names and uses the red/blue color to determine nature. All neutral natures are treated the same because of this.

Tested in english only, other languages were all copied from a bulbapedia table.

### DIFF
--- a/SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxNatureDetector.cpp
+++ b/SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxNatureDetector.cpp
@@ -1,11 +1,10 @@
-/*  IV Checker Reader
+/*  Box Nature Detector
  *
  *  From: https://github.com/PokemonAutomation/Arduino-Source
  *
  */
 
 #include "CommonFramework/ImageTypes/ImageViewRGB32.h"
-#include "PokemonSwSh/Inference/PokemonSwSh_IVCheckerReader.h"
 #include "PokemonBDSP_BoxNatureDetector.h"
 
 namespace PokemonAutomation{
--- a/SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxNatureDetector.cpp
+++ b/SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxNatureDetector.cpp
@@ -1,31 +1,32 @@
-/*  Box Nature Detector
+/*  IV Checker Reader
  *
  *  From: https://github.com/PokemonAutomation/Arduino-Source
  *
  */
 
 #include "CommonFramework/ImageTypes/ImageViewRGB32.h"
-#include "PokemonSV_BoxNatureDetector.h"
+#include "PokemonSwSh/Inference/PokemonSwSh_IVCheckerReader.h"
+#include "PokemonBDSP_BoxNatureDetector.h"
 
 namespace PokemonAutomation{
 namespace NintendoSwitch{
-namespace PokemonSV{
+namespace PokemonBDSP{
 
-const NatureReader& NATURE_READER(){
-    const static Pokemon::NatureReader reader("PokemonSV/NatureCheckerOCR.json");
+const NatureReader& NATURE_READER() {
+    const static Pokemon::NatureReader reader("Pokemon/NatureCheckerOCR.json");
     return reader;
 }
 
 BoxNatureDetector::BoxNatureDetector(VideoOverlay& overlay, Language language)
     : m_language(language)
-    , m_box_nature(overlay, {0.757, 0.529, 0.141, 0.050})
+    , m_nature(overlay, {0.785, 0.196 + 6 * 0.0529, 0.2, 0.0515})
 {}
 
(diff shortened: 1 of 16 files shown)


SerialPrograms/Source/PokemonBDSP/Inference/BoxSystem/PokemonBDSP_BoxNatureDetector.cpp
```diff
@@ -0,0 +1,48 @@
+/*  IV Checker Reader
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include "CommonFramework/ImageTypes/ImageViewRGB32.h"
+#include "PokemonSwSh/Inference/PokemonSwSh_IVCheckerReader.h"
```
Mysticial, line 8: Is this include needed?

kichithewolf: forgot to remove it, removed with latest commit

SerialPrograms/Source/PokemonSwSh/Inference/PokemonSwSh_BoxNatureDetector.cpp
```diff
@@ -0,0 +1,101 @@
...
+    const size_t num_red_pixels = image_stats(red_region).count;
+
+    // Retain only blue pixels from region
+    const ImageRGB32 blue_region = filter_rgb32_range(
+        region,
+        combine_rgb(0, 0, 180), combine_rgb(130, 130, 255), Color(0), replace_color_within_range
+    );
+    const size_t num_blue_pixels = image_stats(blue_region).count;
+
+    const double threshold = region.width() * region.height() * m_area_ratio_threshold;
```
Mysticial, line 52: Where is `m_area_ratio_threshold` being initialized?

SerialPrograms/Source/PokemonSwSh/Inference/PokemonSwSh_BoxNatureDetector.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+BoxNatureDetector::BoxNatureDetector(VideoOverlay& overlay)
+    : m_box_atk(overlay, { 0.689, 0.198 + 1 * 0.0515, 0.84, 0.047})
+    , m_box_def(overlay, { 0.689, 0.198 + 2 * 0.0515, 0.84, 0.047})
+    , m_box_spatk(overlay, { 0.689, 0.198 + 3 * 0.0515, 0.84, 0.047})
+    , m_box_spdef(overlay, { 0.689, 0.198 + 4 * 0.0515, 0.84, 0.047})
+    , m_box_spd(overlay, { 0.689, 0.198 + 5 * 0.0515, 0.84, 0.047})
+{}
+
+NaturePlusMinus BoxNatureDetector::read(Logger& logger, const ImageViewRGB32& frame, const OverlayBoxScope& box){
```
Mysticial, line 33: Nit: You can just take `ImageFloatBox` as the type instead of `OverlayBoxScope`.

SerialPrograms/Source/PokemonSwSh/Inference/PokemonSwSh_BoxNatureDetector.cpp
```diff
@@ -0,0 +1,101 @@
...
+    const ImageRGB32 red_region = filter_rgb32_range(
+        region,
+        combine_rgb(150, 0, 0), combine_rgb(255, 100, 100), Color(0), replace_color_within_range
+    );
+    const size_t num_red_pixels = image_stats(red_region).count;
+
+    // Retain only blue pixels from region
+    const ImageRGB32 blue_region = filter_rgb32_range(
+        region,
+        combine_rgb(0, 0, 180), combine_rgb(130, 130, 255), Color(0), replace_color_within_range
```
Mysticial, line 48: Can you check to make sure this works on both of these test images with room to spare on the thresholds? These are two different capture cards, one of which is has significant color distortion.
screenshot-20230128-010949517344

Probably want to ping Gin later on to add these to the test unit tests.

screenshot-20230128-010921021655

SerialPrograms/Source/PokemonSwSh/Inference/PokemonSwSh_BoxNatureDetector.cpp
```diff
@@ -23,49 +23,26 @@ namespace NintendoSwitch{
...
+    //cout << region.average.r << endl;
+    //cout << region.average.g << endl;
+    //cout << region.average.b << endl;
+    
+    if (region.average.b > region.average.r + 3) {
+        return NaturePlusMinus::MINUS;
     }
-
-    if (num_red_pixels > threshold) {
+    else if (region.average.r > region.average.b) {
```
Mysticial, line 43: Is this actually reliable? There's only a 3 pixel window that can return neutral. It's easy for a capture cards to tint slightly in one direction.

What are the actual average RGB values you get for the different test images? When it's red, is the red much larger than the blue? Perhaps you can filter out the white background first, then average only the pixels that are left.

PR #266, opened 2023-01-27
new option to select start location for chest farmer, handle chi-yu using bounce

kichithewolf:
PR with updated instructions is also open in CC repo

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
@@ -146,6 +146,7 @@ void StatsReset::program(SingleSwitchProgramEnvironment& env, BotBaseContext& co
             throw OperationFailedException(env.console, "Failed to enter battle. Are you facing the Pokemon or in a menu?", true);
         }
         bool battle_ended = false;
+        bool goldfish_loop = false;
         while (!battle_ended){
             //Navigate to correct ball and repeatedly throw it until caught
             pbf_press_button(context, BUTTON_X, 20, 100);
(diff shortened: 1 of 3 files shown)


SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
```diff
@@ -166,27 +167,46 @@ void StatsReset::program(SingleSwitchProgramEnvironment& env, BotBaseContext& co
...
             pbf_mash_button(context, BUTTON_A, 150);
-            stats.balls++;
-            env.update_stats();
-            //pbf_wait(context, 4000);
-            pbf_mash_button(context, BUTTON_B, 1000);
-            context.wait_for_all_requests();
 
-            AdvanceDialogWatcher summary(COLOR_MAGENTA);
-            int ret2 = wait_until(
+            //Check for battle menu - if found after a second then assume caught in a goldfish bounce loop
```
pifopi, line 173: Should it be 4 or 1 second ?

pifopi (approved):

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
Mysticial: I think the comment is wrong. 4 is correct since it takes a lot longer than 4 seconds from throwing a ball to returning to the battle menu after breaking out. Whereas if you are stuck, you see it a lot faster - but possibly longer than 1 second depending on text speed and how laggy the game is.

Mysticial: I'll merge and fix the comment in a followup.

PR #267, opened 2023-01-28
Replace Sleepy with D++

Koi-3088:


Koi-3088:
Please play around and test before merging/rebasing.

Mysticial:
What changed with the two networking DLLs?

- libcrypto-1_1-x64.dll
- libssl-1_1-x64.dll

Mysticial:
Can we make this compile for Mac and Linux as well, but with DPP disabled? (like the case for Sleepy right now)

Mysticial:
The big topic here is how we actually migrate from Sleepy to DPP.

What we have right now is reasonably stable (with Sleepy) and a lot of people are using it. Since the DPP implementation is new, it's bound to have breaking issues.

So the problem with removing Sleepy and adding DPP in a single big-bang change is that we go from the stable Sleepy implementation to the unstable DPP one for an otherwise in-production product which will probably break it for some time until we fix all the bugs.

So I'm thinking of two alternative approaches for this migration - neither of which are great:

1. Don't remove Sleepy yet. Have both Sleepy and DPP together with a compiler or settings option to switch between them. That we can test and flesh out DPP before we switch everyone over and remove Sleepy.

2. Maintain a separate branch that does this big-bang change. Then we release the DPP binary into internal/beta and iterate on that until it is ready. But this will involve a lot of potentially messy rebases against main.

Koi-3088:
DLLs probably just came with DPP, I'm not sure if they're any different aside from size.

I can try to go with the first approach and `ifdef` both Sleepy and DPP. Thanks for the feedback!

SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
```diff
@@ -0,0 +1,155 @@
...
+                            msg += "<@" + (std::string)settings.message.user_id + ">";
+                        }
+
+                        const std::string& discord_message = settings.message.message;
+                        if (!discord_message.empty()) {
+                            if (!msg.empty()) {
+                                msg += " ";
+                            }
+
+                            for (char ch : discord_message) {
```
pifopi, line 122: doing this isn't great for performance reason. You are resizing the string a lot.

discord_message.erase(std::remove(discord_message.begin(), discord_message.end(), "@"), discord_message.end());
msg += discord_message;

plays better with memory allocation because you resize only once

pifopi (approved):
looks good to me !

SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
Koi-3088: I honestly just copied the embed building from Sleepy, not even sure if it was mine or Kuro's 😆

SerialPrograms/Source/CommonFramework/Notifications/ProgramNotifications.cpp
```diff
@@ -182,14 +180,14 @@ void send_raw_notification(
     );
 #ifdef PA_SLEEPY
     Integration::SleepyDiscordRunner::send_message_sleepy(
```
pifopi, line 182: there should be a
If (GlobalSettings::instance().DISCORD.integration.run_on_start && !GlobalSettings::instance().DISCORD.integration.use_dpp)

otherwise, the first move will nuke the object and dpp won't shend shit

SerialPrograms/Source/CommonFramework/Notifications/ProgramNotifications.cpp
```diff
@@ -232,14 +228,14 @@ void send_raw_notification(
     );
 #ifdef PA_SLEEPY
     Integration::SleepyDiscordRunner::send_message_sleepy(
-        should_ping, tags, "", std::move(embed_integration),
+        should_ping, tags, "", std::move(embed),
```
pifopi, line 231: same here

pifopi (approved):

SerialPrograms/Source/Integrations/DppIntegration/DppUtility.cpp
```diff
@@ -0,0 +1,87 @@
...
+                    count += map.size();
+                }
+
+                user_counts.emplace(id, count);
+                log("User count: " + std::to_string(count) + " (" + event.created->name + ")", "get_user_counts()", ll_info);
+            }
+
+            uint64_t Utility::sanitize_console_id_input(const slashcommand_t& event) {
+                int64_t id = std::get<int64_t>(event.get_parameter("id"));
+                if (id > UINT64_MAX) {
```
Mysticial, line 64: I'm surprised you're not getting a compiler warning here about comparing signed and unsigned integer.

This if-statement will never return true since `id` will be cast to `uint64_t` first. Then no `uint64_t` can be greater than `UINT64_MAX`.

Are you trying to check for negative `id`?

Koi-3088: I'm actually not sure why I still have this method here, I had it before I learned I can enforce min values in slash commands, and then it did have a negative check. Will yeet in just a moment.

3rdParty/dpp/nlohmann/json_fwd.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <memory> // allocator
+#include <string> // string
+#include <vector> // vector
+
+/*!
+@brief namespace for Niels Lohmann
+@see https://github.com/nlohmann
+@since version 1.0.0
+*/
+namespace nlohmann
```
Mysticial, line 15: No namespace conflict or ODR violation with `3rdParty/nlohmann/json.hpp`?

Koi-3088: Not as far as I can tell. Not getting any compiler warnings, nlohmann seems to be accessible and usable just fine.

SerialPrograms/Source/Integrations/DppIntegration/DppCommandHandler.cpp
```diff
@@ -0,0 +1,399 @@
...
+                    }
+                }
+
+                if (!msg.empty() && msg != "") {
+                    m.content = msg;
+                }
+
+                m.channel_id = channel;
+                m.add_embed(embed);
+                bot.message_create(m);
```
Mysticial, line 110: This is DPP's send command? haha

Koi-3088: Yeah, though depends on context.

SerialPrograms/Source/Integrations/DppIntegration/DppCommandHandler.cpp
```diff
@@ -0,0 +1,399 @@
...
+                            it->second.func(event);
+                        }
+                        catch (std::exception& e) {
+                            this->log_dpp("Exception occurred while executing command: " + event.command.get_command_name() + ".\n" + e.what(), "Slash Command", ll_critical);
+                        }
+                    }
+                });
+            }
+
+            void Handler::send_message(cluster& bot, embed& embed, const std::string& channel, std::chrono::milliseconds delay, const std::string& msg, std::shared_ptr<PendingFileSend> file) {
```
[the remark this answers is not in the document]
Koi-3088: No, I passed the argument but haven't implemented the `ScheduledTaskRunner`. Mainly because I forgot, but DPP also sends asynchronously, so not sure if as necessary as with Sleepy. Though can't hurt.

3rdParty/dpp/nlohmann/json_fwd.hpp
Mysticial: The ODR part scares me a bit. Since the linker can mix things up across the two files. We're probably getting lucky because they're only ever being used in two completely separate contexts.

I'll merge, then change the namespace of our existing nlohmann as the short-term fix. The proper solution is to point both CC and DPP to the same nlohmann file. But that could be a bit messy.

PR #271, opened 2023-01-30
Fix user counts?

Koi-3088:
Also "instant" exit.

### DIFF
--- a/SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
+++ b/SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
@@ -27,16 +27,18 @@ bool Client::is_running() {
 }
 
 void Client::connect() {
-    std::lock_guard<std::mutex> lg(m_connect_lock);
+    std::lock_guard<std::mutex> lg(m_client_lock);
     if (m_bot == nullptr && !m_is_connected) {
         DiscordSettingsOption& settings = GlobalSettings::instance().DISCORD;
         if (!Handler::check_if_empty(settings))
             return;
 
         std::string token = settings.integration.token;
+        uint32_t intents = intents::i_default_intents | intents::i_guild_members;
         try {
+            m_bot = std::make_unique<cluster>(token, intents);
+            m_bot->cache_policy = { cache_policy_setting_t::cp_lazy, cache_policy_setting_t::cp_lazy, cache_policy_setting_t::cp_aggressive };
             std::thread(&Client::run, this, token).detach();
-            m_is_connected = true;
         }
         catch (std::exception& e) {
             Handler::log_dpp("DPP thew an exception: " + (std::string)e.what(), "connect()", ll_critical);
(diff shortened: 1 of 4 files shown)


SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
```diff
@@ -152,11 +151,14 @@ void Client::run(const std::string& token) {
     try {
         uint32_t intents = intents::i_default_intents | intents::i_guild_members;
         m_bot = std::make_unique<cluster>(token, intents);
```
Mysticial, line 153: This still has the problem of initializing asynchronously from the starting thread. Likewise, if `connect()` is called very quickly twice in the row before the thread can start up, it will crash.

Is it possible to move this line `m_bot = std::make_unique<cluster>(token, intents);` to line 37 (just before the thread is started)?

This also protects against multiple `connect()` calls since the first one will get in and construct it. Then the later ones will see that is has already been constructed and will exit.

SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
```diff
@@ -152,11 +151,14 @@ void Client::run(const std::string& token) {
...
+        m_bot->cache_policy = { cache_policy_setting_t::cp_lazy, cache_policy_setting_t::cp_lazy, cache_policy_setting_t::cp_aggressive };
         Handler::initialize(*m_bot.get());
-        m_bot->start(false);
+        m_bot->start(st_return);
+        m_is_connected = true;
     }
     catch (std::exception& e) {
         Handler::log_dpp("DPP thew an exception: " + (std::string)e.what(), "run()", ll_critical);
+        m_bot.reset();
```
Mysticial, line 161: This is also dangerous. Though I think it can be fixed by covering the entire function with the lock.

SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
Koi-3088: It's possible for the `start` button to be clicked in rapid succession? Yeah, it can definitely be moved.

Mysticial: Doesn't necessarily have to be that quick. Could also be CPU starved such at the new thread takes forever to startup. At this is the obvious example of how it can get called twice. There might be more involving how it's auto-started with the program.

Either way, better to make it robust to this since someone may call it differently in the future. lol

SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
```diff
@@ -118,14 +121,15 @@ void Client::send_embed_dpp(
 }
 
 void Client::run(const std::string& token) {
+    std::lock_guard<std::mutex> lg(m_client_lock);
```
Mysticial, line 124: After merging this, I'll add a null check for `m_bot` here. It can be null if `disconnect()` gets called immediately after `connect()`, but before the thread can start up.

`connect()` - Constructs `m_bot` and creates the thread.
`disconnect()` - Destructs `m_bot` and sets it back to null.
`run()` - Thread finally starts running and sees that `m_bot` is null.

PR #275, opened 2023-01-31
Tera Self Farmer Herba Mystica Filter

Vaedz7:
- Doesn't interfere with min/max stars skipping raid
- If raid is found within given range, the filter would only battle Pokemon with high herba drop rates


SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraSelfFarmer.cpp
```diff
@@ -344,6 +351,43 @@ void TeraSelfFarmer::program(SingleSwitchProgramEnvironment& env, BotBaseContext
             close_raid(env.program_info(), env.console, context);
             continue;
         }
+        
+        const char* sixstar[6]
```
pifopi, line 355: I think the code would be cleaner if you moved that to a specific function !

Also, i strongly advise to use std::array + for each. SOmething that would looks like this if you're not familiar with it :



    std::array<std::string, 6> sixstar{"blissey", "vaporeon", "amoonguss", "farigiraf", "cetitan", "dondozo"};
    std::array<std::string, 9> fivestar{"gengar", "glalie", "amoonguss", "dondozo", "palafin", "blissey", "eelektross", "driftblim", "cetitan"};
    
    for (const std::string& str : sixstar){
        if (best_silhouette == str){
             return false;
        }
    }

pifopi: that way, there is way less chance for anyone to do a mistake when editing the code, etc

PR #280, opened 2023-02-01
Focus Herba Option

Vaedz7:


Mysticial:
Seeing how big this is getting, let's start by moving the `TeraFarmerOpponentFilter` class into a separate file.

After that, we need to think about the big picture here. Here you'd adding so many special cases for the various herbs that it's raising these questions:
1. How many people are actually going to use it?
2. How will it conflict with future filters?

The big potential future is a whitelist/blacklist table where you can specify to either do/skip raids based on species, tera type, stars, etc... Basically similar to the Egg Autonomous table. In this case, it actually makes more sense for these herba-targeting filters to saved tables json files which people can load instead of being hard-coded like this.

Vaedz7:
With the new sandwiches, if we want to move the filter into a new file it’s probably not worth it as the amount of people using specific herbs will be less. The only use case I can honestly think of is for sweet herba for eggs

Vaedz7:
Maybe just keep the filter as it is at the moment with only the general Herba Option to keep the filter for becoming large and complicated for no real reason

blake-nx:
@Vaedz7 Can we add Finizen to this fivestar list? 

Finizen doesn't show up in fivestar raids, but it has the same silhouette as Palafin, so SKIP_HERBA is misidentifying Palafin as Finizen and skipping it.

Finizen silhouette
2023-02-06_08h50_40
Palafin silhouette
2023-02-06_12h52_16

Palafin identified as Finizen
2023-02-06_08h07_50
2023-02-06_08h04_44

PR #282, opened 2023-02-04
- Add a unified command handler to make use of both slash and prefix

Koi-3088:
…commands.

- Add option to use either slash commands or prefix commands.
- Replace `std::format()` for cross-compatibility.

### DIFF
--- a/SerialPrograms/Source/Integrations/DppIntegration/DppUtility.cpp
+++ b/SerialPrograms/Source/Integrations/DppIntegration/DppUtility.cpp
@@ -58,6 +58,27 @@ uint16_t Utility::get_button(const uint16_t& bt) {
     return 1 << bt;
 }
 
+int64_t Utility::get_value_from_input(const commandhandler& handler, const std::string& command_name, const std::string& input, std::string& out) {
+    auto cmd = handler.commands.find(command_name);
+    auto& choices = cmd->second.parameters[1].second.choices;
+    for (auto& choice : choices) {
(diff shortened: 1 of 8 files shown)


Koi-3088:
Pain.

SerialPrograms/Source/Integrations/DppIntegration/DppUtility.cpp
```diff
@@ -58,6 +58,27 @@ uint16_t Utility::get_button(const uint16_t& bt) {
     return 1 << bt;
 }
 
+int64_t Utility::get_value_from_input(const commandhandler& handler, const std::string& command_name, const std::string& input, std::string& out) {
+    auto cmd = handler.commands.find(command_name);
+    auto& choices = cmd->second.parameters[1].second.choices;
```
Mysticial, line 63: No chance `cmd->second.parameters` has fewer than 2 elements right?

Koi-3088: `commandhandler` expects you to know the size and parameter types. If you supply empty parameters, it will fail the `parameter.size()` check. If you provide at least one parameter, the rest get initialized as defaults. In other words, this should (surprisingly) be safe.

Koi-3088: In this particular case, I'm iterating through what we have generated ourselves, so it wouldn't fail anyways. Bu, again, we're expected to know this.

PR #285, opened 2023-02-05
Tournament farmer

kichithewolf:
Farms the academy tournament and notifies if the resulting prize matches filters. Supports tera at the start of each battle and saving every x number of tournaments. If tournament lost, it will attempt to return to the academy. Does not support counting prize money (yet). Turbo A actually runs faster than this if you have the right build, don't die, and don't care for knowing what the prizes were so I'll put that in the documentation when I get to writing it.

Prize reading will most likely have the same gold bottle cap issue in japanese as auction farmer. In english, "a bottle of PP Up" was mistaken as "bottle cap" so there's another line in the json file to account for it, unsure if a similar issue will happen in other languages. I couldn't find translations for "tera shard" so it only checks for the type.

### DIFF
--- a/SerialPrograms/Source/CommonFramework/OCR/OCR_Routines.cpp
+++ b/SerialPrograms/Source/CommonFramework/OCR/OCR_Routines.cpp
@@ -97,6 +97,14 @@ const std::vector<TextColorRange>& BLACK_OR_WHITE_TEXT_FILTERS(){
     };
     return filters;
 }
+const std::vector<TextColorRange>& BLUE_TEXT_FILTERS() {
+    static std::vector<TextColorRange> filters{
+        {0xFF3287A2, 0xFFA7F4FF},
+        {0xFF2283A2, 0xFF8EF1FF},
+        {0xFF428BA2, 0xFFC1F7FF},
+    };
+    return filters;
+}
 
 
 
--- a/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.cpp
+++ b/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.cpp
@@ -1,0 +1,39 @@
+/*  Tournament Prize Reader
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include "CommonFramework/OCR/OCR_RawOCR.h"
+#include "PokemonSV_TournamentPrizeNameReader.h"
+
+namespace PokemonAutomation {
+namespace NintendoSwitch {
+namespace PokemonSV {
+
+TournamentPrizeNameReader& TournamentPrizeNameReader::instance() {
+    static TournamentPrizeNameReader reader;
+    return reader;
+}
+
+
+TournamentPrizeNameReader::TournamentPrizeNameReader()
+    : SmallDictionaryMatcher("PokemonSV/AAT/TournamentPrizeNameOCR.json")
+{}
+
+OCR::StringMatchResult TournamentPrizeNameReader::read_substring(
+    Logger& logger,
+    Language language,
+    const ImageViewRGB32& image,
+    const std::vector<OCR::TextColorRange>& text_color_ranges,
+    double min_text_ratio, double max_text_ratio
+) const {
+    return match_substring_from_image_multifiltered(
+        &logger, language, image, text_color_ranges,
+        MAX_LOG10P, MAX_LOG10P_SPREAD, min_text_ratio, max_text_ratio
+    );
+}
+
+}
+}
+}
--- a/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.cpp
+++ b/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.cpp
@@ -1,4 +1,4 @@
-/*  Tournament Prize Reader
+/*  Tournament Prize Name Reader
  *
  *  From: https://github.com/PokemonAutomation/Arduino-Source
  *
--- a/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.h
+++ b/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.h
@@ -1,11 +1,11 @@
-/*  Tournament Prize Reader
+/*  Tournament Prize Name Reader
  *
  *  From: https://github.com/PokemonAutomation/Arduino-Source
  *
  */
 
-#ifndef PokemonAutomation_PokemonSV_TournamentPrizeReader_H
-#define PokemonAutomation_PokemonSV_TournamentPrizeReader_H
+#ifndef PokemonAutomation_PokemonSV_TournamentPrizeNameReader_H
+#define PokemonAutomation_PokemonSV_TournamentPrizeNameReader_H
 
 #include "CommonFramework/OCR/OCR_SmallDictionaryMatcher.h"
 
--- a/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.h
+++ b/SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.h
@@ -1,0 +1,38 @@
+/*  Tournament Prize Reader
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#ifndef PokemonAutomation_PokemonSV_TournamentPrizeReader_H
+#define PokemonAutomation_PokemonSV_TournamentPrizeReader_H
+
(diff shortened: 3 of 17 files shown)


SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.cpp
```diff
@@ -0,0 +1,39 @@
+/*  Tournament Prize Reader
```
pifopi, line 1: add Name

SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.h
```diff
@@ -0,0 +1,38 @@
+/*  Tournament Prize Reader
```
pifopi, line 1: add Name

SerialPrograms/Source/PokemonSV/Inference/PokemonSV_TournamentPrizeNameReader.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#ifndef PokemonAutomation_PokemonSV_TournamentPrizeReader_H
```
pifopi, line 7: add Name

SerialPrograms/Source/PokemonSV/Options/PokemonSV_TournamentPrizeSelectOption.cpp
```diff
@@ -0,0 +1,52 @@
+/*  Tournament Prize Select
```
pifopi, line 1: add Option

SerialPrograms/Source/PokemonSV/Options/PokemonSV_TournamentPrizeSelectOption.h
```diff
@@ -0,0 +1,24 @@
+/*  Tournament Prize Select
```
pifopi, line 1: add Option

SerialPrograms/Source/PokemonSV/Options/PokemonSV_TournamentPrizeTable.h
```diff
@@ -0,0 +1,51 @@
+/* Tournament prize table
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#ifndef PokemonAutomation_PokemonSV_TournamentPrizeSelector_H
```
pifopi, line 7: Table

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -0,0 +1,370 @@
...
+    find prize sprites - some code is there, just disabled
+    find translations for tera shards
+    other languages: make sure "bottle cap" isn't misread as "bottle of PP Up"
+    */
+
+    for(uint32_t c = 0; c < NUM_ROUNDS; c++) {
+        //Initiate dialog then mash until first battle starts
+        AdvanceDialogWatcher advance_detector(COLOR_YELLOW);
+        pbf_press_button(context, BUTTON_A, 10, 50);
+        pbf_press_button(context, BUTTON_A, 10, 50); //sometimes the input drops
```
pifopi, line 135: is it only the first time ? or every time (the first button press is usually lost connecting the controller)

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+            int ret_battle2 = run_until(
+                env.console, context,
+                [](BotBaseContext& context) {
+                    pbf_mash_button(context, BUTTON_B, 4000);
+                },
+                { battle_menu2, overworld }
+                );
+            context.wait_for_all_requests();
+            int ret_black;
```
pifopi, line 171: move inside block

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -0,0 +1,370 @@
...
+            { battle_menu }
+            );
+        if (ret_battle != 0) {
+            env.console.log("Failed to detect battle start!", COLOR_RED);
+        }
+        context.wait_for_all_requests();
+
+        BlackScreenWatcher end_of_battle(COLOR_RED, { 0.2, 0.2, 0.6, 0.6 });
+        bool battle_lost = false;
+        for (uint16_t battles = 0; !battle_lost && battles < 4; battles++) {
```
pifopi, line 158: doing a break if battle lost seems cleaner to me. It's easy to miss it here

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -0,0 +1,370 @@
...
+    PA_ADD_OPTION(NUM_ROUNDS);
+    PA_ADD_OPTION(TRY_TO_TERASTILLIZE);
+    PA_ADD_OPTION(SAVE_NUM_ROUNDS);
+    PA_ADD_OPTION(GO_HOME_WHEN_DONE);
+    PA_ADD_OPTION(LANGUAGE);
+    PA_ADD_OPTION(TARGET_ITEMS);
+    PA_ADD_OPTION(NOTIFICATIONS);
+}
+
+void TournamentFarmer::program(SingleSwitchProgramEnvironment& env, BotBaseContext& context) {
```
pifopi, line 113: could you break that function into smaller blocks ? Like one block for one fight, one block for battle lost management, etc ? It'd be easier to understand what's going on

pifopi (approved):

SerialPrograms/Source/CommonFramework/OCR/OCR_Routines.cpp
```diff
@@ -97,6 +97,14 @@ const std::vector<TextColorRange>& BLACK_OR_WHITE_TEXT_FILTERS(){
     };
     return filters;
 }
+const std::vector<TextColorRange>& BLUE_TEXT_FILTERS() {
+    static std::vector<TextColorRange> filters{
+        {0xFF3287A2, 0xFFA7F4FF},
+        {0xFF2283A2, 0xFF8EF1FF},
+        {0xFF428BA2, 0xFFC1F7FF},
```
Mysticial, line 104: Were these found empirically? Or copied from somewhere?

kichithewolf: Took a screenshot in-program and found the color value used, then went 20% darker, 20% lighter. The two other lines are more/less saturation.

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -0,0 +1,412 @@
...
+            }
+        }
+
+        //Tournament won
+        if (!battle_lost) {
+            handle_end_of_tournament(env, context);
+        }
+
+        //Save the game if option is set
+        if (SAVE_NUM_ROUNDS != 0 && ((c + 1) % SAVE_NUM_ROUNDS) == 0) {
```
Mysticial, line 399: `SAVE_NUM_ROUNDS` is a config option that's unlocked and can change at any time. So it's possible for it to be non-zero first, thus passing `SAVE_NUM_ROUNDS != 0`. Then it changes to zero before it does the `% SAVE_NUM_ROUNDS` which will lead to a crash.

Read it once and store to a variable before running this if-statement. That way both parts of the if-statement will be guaranteed to check the same value.

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -0,0 +1,412 @@
...
+        pbf_press_button(context, BUTTON_R, 20, 50);
+    }
+
+    //Mash A until battle finished
+    BlackScreenWatcher end_of_battle(COLOR_RED, { 0.2, 0.2, 0.6, 0.6 });
+    int ret_black = run_until(
+        env.console, context,
+        [](BotBaseContext& context) {
+            pbf_mash_button(context, BUTTON_A, 30000);
+            pbf_mash_button(context, BUTTON_A, 7500); //5 minutes should be more than enough for one battle
```
Mysticial, line 136: This kind of A-mashing over such a long duration is a bit risky. If the game crashes and there's an update available, it will run the update.

A better approach here is to search for check for the battle menu, and only mash A for a couple seconds at a time when you see the battle menu. IOW, add `NormalBattleMenuDetector` as an additional callback here to tell you when you press A.

PR #300, opened 2023-02-16
[SV MC] Adding EggHatcher & Updating EggFetcher

nymphaea-hydriades:


NativePrograms/PokemonSV/Programs/EggCombined_Core.c
```diff
@@ -0,0 +1,205 @@
...
+#include "NativePrograms/DeviceFramework/DeviceSettings.h"
+#include "NativePrograms/NintendoSwitch/Libraries/FrameworkSettings.h"
+#include "NativePrograms/NintendoSwitch/Libraries/NintendoSwitch_Device.h"
+#include "NativePrograms/NintendoSwitch/Libraries/NintendoSwitch_PushButtons.h"
+#include "EggCombined.h"
+
+void open_picnic(void) {
+    pbf_move_left_joystick(STICK_MIN, STICK_CENTER, 190, 0);
+    pbf_press_button(BUTTON_L, 5, 100);
+    pbf_press_button(BUTTON_X, 1, 200);
```
Mysticial, line 11: I'm surprised this even works. Button presses held for fewer than 5 ticks are generally not even recognized by the Switch.

Has this been tested? For SV, I generally hold for 20 ticks whereas for SwSh the minimum of 5 is enough.

Mysticial: Oh I see. Internally, `pbf_press_button()` increases it to 5 if it's below 5. Probably not a behavior we should rely on.

Let's increase all these small hold durations to at least 5. (ideally 20 or more for SV)

nymphaea-hydriades: Ah that makes sense, I saw some of the other programs had 1 tick presses and nothing strange happened during test runs so didn't think too much of it. Let me extend the press durations for all of them and redo the test runs to adjust timing, I'll get back to you once that's done.

nymphaea-hydriades: Button press durations are now extended to be 20 ticks for most cases. 5 ticks are used when outside of the game, spamming DPAD during box operations, and attempting to dash during hatching. Thank you for this comment.

PR #303, opened 2023-02-24
tournament money tracking and options

kichithewolf:
tournament farmer now reads and tracks money earned. new option added to stop after x amount of money made.

read_money is a modified read_number with some regex. depending on the lighting/time of day in game, additional digits would be added when trying to read the bottom notification. read_money stops that by only taking the amount after £


kichithewolf:
last commit is because this image is 82.5114
20230224-115151749280-OperationFailedException

PR #305, opened 2023-02-25
tournament loss fix, improve logging

kichithewolf:


SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_TournamentFarmer.cpp
```diff
@@ -505,6 +508,34 @@ void TournamentFarmer::program(SingleSwitchProgramEnvironment& env, BotBaseConte
                 break;
             }
 
+            //If this is the last battle in the tournament check for overworld in case player lost
+            if (battles == 3) {
+                env.log("Final battle of the tournament complete, checking for overworld/loss.");
+
+                //Clear dialog, mash B
+                pbf_mash_button(context, BUTTON_B, 400);
```
Mysticial, line 516: Is there any risk of mashing B so far that you crouch in the overworld?

kichithewolf: It crouches a few times in front of the pokemon center. I left it as a buffer in case of slower switches.

PR #308, opened 2023-02-27
implement SystemCpuTime for macOS/Linux

naussika:
so now CPU Utilization works
Screenshot 2023


pifopi:
It seems to fail to compile with :
https://github.com/PokemonAutomation/Arduino-Source/actions/runs/4283679056/jobs/7459677758#step:10:113
Could you fix that ?

PR #312, opened 2023-03-05
EventNotificationOption::parse_tags: don't use QT

naussika:


SerialPrograms/Source/CommonFramework/Notifications/EventNotificationOption.cpp
```diff
@@ -85,14 +85,14 @@ std::string EventNotificationOption::tags_to_str(const std::vector<std::string>&
...
-        if (!cstr.empty()){
-            tags.emplace_back(std::move(cstr));
+    for (size_t start = 0, end = 0; (end = str.find(",", start)) != std::string::npos; start = end + 1){
+        std::string token = str.substr(start, end - start);
+        token = sanitize_tag(token);
+        if (!token.empty()){
+            tags.emplace_back(std::move(token));
         }
     }
+
```
Mysticial, line 95: I think this misses the last token since it will be missing the comma.

PR #342, opened 2023-07-05
additional checks for caught or fainted pokemon

kichithewolf:
updated logging+notifications, checks for overworld and swap menu in battle loop, checks for correct no. of pokemon before IV read


SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
```diff
@@ -195,18 +197,38 @@ void StatsReset::program(SingleSwitchProgramEnvironment& env, BotBaseContext& co
...
                 AdvanceDialogWatcher summary(COLOR_MAGENTA);
+                OverworldWatcher overworld(COLOR_BLUE);
+                SwapMenuWatcher swap_menu(COLOR_YELLOW);
                 int ret2 = wait_until(
                     env.console, context,
                     std::chrono::seconds(25),
-                    { summary, battle_menu }
+                    { summary, overworld, battle_menu, swap_menu }
                 );
                 if (ret2 == 0) {
```
Mysticial, line 207: There's probably enough cases here to use a switch statement.

PR #344, opened 2023-07-12
Auto-Host Multi-Game

Vaedz7:
New Program

Features
- Runs up to 16 Tera Raids on one switch using different games/profiles
- Opens each game & user instance and runs AutoHost like normal
- Includes a Post-Raid Delay

Notes
- Does not include separate Battle AI's for each game. (I recommend just spamming a move as you can rearrange moves in SV to match positions)
- Does not include separate hosting modes (I'll probably add this in the future)
- Does not include separate settings (Like Lobby Wait, Description, and Start Players; I'll probably add this as well later on)

Overall
This is a very basic AutoHost program to run multiple raids with the same settings. I'm planning to add more as time goes on.

### DIFF
--- a/SerialPrograms/SerialPrograms.pro
+++ b/SerialPrograms/SerialPrograms.pro
@@ -710,6 +710,7 @@ SOURCES += \
     Source/PokemonSV/Programs/ShinyHunting/PokemonSV_LetsGoTools.cpp \
     Source/PokemonSV/Programs/ShinyHunting/PokemonSV_ShinyHunt-AreaZeroPlatform.cpp \
     Source/PokemonSV/Programs/TeraRaids/PokemonSV_AutoHost.cpp \
+    Source/PokemonSV/Programs/TeraRaids/PokemonSV_AutoHost-MultiGame.cpp \
     Source/PokemonSV/Programs/TeraRaids/PokemonSV_AutoHostLobbyWaiter.cpp \
     Source/PokemonSV/Programs/TeraRaids/PokemonSV_AutoHostTools.cpp \
     Source/PokemonSV/Programs/TeraRaids/PokemonSV_JoinTracker.cpp \
@@ -924,7 +925,6 @@ SOURCES += \
     Source/ZeldaTotK/Programs/ZeldaTotK_MineruItemDuper.cpp \
     Source/ZeldaTotK/Programs/ZeldaTotK_ParaglideItemDuper.cpp \
     Source/ZeldaTotK/Programs/ZeldaTotK_SurfItemDuper.cpp \
-    Source/ZeldaTotK/Programs/ZeldaTotK_WeaponDuper.cpp \
     Source/ZeldaTotK/ZeldaTotK_Panels.cpp \
     Source/ZeldaTotK/ZeldaTotK_Panels.h \
     Source/ZeldaTotK/ZeldaTotK_Settings.cpp \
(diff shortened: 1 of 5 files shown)


pifopi:
you should rebease your PR. The link issues are because my changes are not present on your PR, sorry about that

Vaedz7:
> you should rebase your PR. The link issues are because my changes are not present on your PR, sorry about that

What changes are those? For me it's showing that the branch is fully updated and when I create a new branch from PokemonAutomation/main it is the same as my PR (excluding my changes).

pifopi:
oh it was actually rebased but this change :
https://github.com/Vaedz7/Arduino-Source/commit/f358d136553d2739b52346f5d7e60a446cff070b
removed some of my changes so it broke the linking ^^

Vaedz7:
> oh it was actually rebased but this change : [Vaedz7@f358d13](https://github.com/Vaedz7/Arduino-Source/commit/f358d136553d2739b52346f5d7e60a446cff070b) removed some of my changes so it broke the linking ^^

Oh ok! I've fixed that now!

Mysticial:
How does this work? If I want to host two dens set to different dates, one of them will be wiped when it tries to go online with the later date? I'm not seeing any date manipulation logic.

Vaedz7:
> How does this work? If I want to host two dens set to different dates, one of them will be wiped when it tries to go online with the later date? I'm not seeing any date manipulation logic.

I now realize that the program should switch dates for dens so I'll add that as an option by tomorrow. I'm not sure if one would be wiped as they are on different games/users so I'll test that out as well.

SerialPrograms/SerialPrograms.pro
```diff
@@ -1961,7 +1964,7 @@ HEADERS += \
     Source/ZeldaTotK/Programs/ZeldaTotK_BowItemDuper.h \
     Source/ZeldaTotK/Programs/ZeldaTotK_MineruItemDuper.h \
     Source/ZeldaTotK/Programs/ZeldaTotK_ParaglideItemDuper.h \
-    Source/ZeldaTotK/Programs/ZeldaTotK_SurfItemDuper.h \
```
pifopi, line 1964: You should keep the \

Vaedz7:
> How does this work? If I want to host two dens set to different dates, one of them will be wiped when it tries to go online with the later date? I'm not seeing any date manipulation logic.

I've now added a ``skips`` option to specify how many days ahead of the current date the den is located. The skips won't need to account for moving ahead a date as AutoHost already takes care of that. I haven't yet tested this but will get to it soon!

Mysticial:
This approach seems prone to error. What about something that lets you specify the date of every den?

Vaedz7:
> This approach seems prone to error. What about something that lets you specify the date of every den?

That may be a better approach. I'll sort that out by next week and test both approaches to see which runs better.

Mysticial:
Looking at the time settings, we only have a setter for the hours. `NintendoSwitch_DateReader.cpp`
For this program we'll also need setters for the month and year. And you'd need to branch it based on the date format.

In short, not a small change. So I suggest we add the full clock manipulation functions as a separate change/PR first. Then come back to this later.

Mysticial:
Also worth mentioning that our date format detection only detects if it's a US format or not. While this is enough to distinguish the hour/minutes/seconds/AMPM format, it's not enough to distinguish the date format which can be either YYYY/MM/DD or DD/MM/YYYY.

Mysticial:
Closing due to inactivity.

PR #346, opened 2023-08-01
Implement new AutoSave mode AutoSave::AfterFetchComplete for sv-egg-autonomous

shinymarill23:
This PR adds a new AutoSave option, `AutoSave::AfterFetchComplete` for the SV egg autonomous program.

Unlike `AutoSave::AfterEveryBatch`, this only saves once: when all eggs have been fetched from the picnic AND before the first batch of hatching begins. This gives the user the ability to "un-hatch" the whole egg box if they want to do so.


pifopi:
Looks good to me !

Gin890:
LGTM!

PR #348, opened 2023-08-03
Add SV Tera Roller

NympheaR:
Faster alternative to roll for a shiny Tera Raid that runs from battle and checks the shininess from the Pokédex Recently Battled section.


SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -398,6 +399,120 @@ void leave_box_system_to_overworld(const ProgramInfo& info, ConsoleHandle& conso
...
+        int ret = wait_until(
+            console, context,
+            std::chrono::seconds(100),
+            {overworld}
+        );
+
+        if (ret == 0){
+            return;
+        } else {
+            env.console.log("leave_phone_to_overworld(): Failed to return to overworld after 100 seconds.");
```
pifopi, line 512: Env is not defined so it break compilation

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -398,6 +399,120 @@ void leave_box_system_to_overworld(const ProgramInfo& info, ConsoleHandle& conso
...
+                ErrorReport::SEND_ERROR_REPORT, console,
+                "open_pokedex_from_overworld(): No overworld state found after 10 seconds.",
+                true
+            );
+        }
+    }
+
+    WallClock start = current_time();
+    while (true){
+        if (current_time() - start > std::chrono::minutes(1)){
```
pifopi, line 427: Could that time be reduced? 1 minute sounds huge

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -398,6 +399,120 @@ void leave_box_system_to_overworld(const ProgramInfo& info, ConsoleHandle& conso
...
+        context.wait_for_all_requests();
+        int ret = wait_until(
+            console, context,
+            std::chrono::seconds(10),
+            {overworld}
+        );
+        context.wait_for(std::chrono::milliseconds(100));
+        if (ret == 0){
+            console.log("Detected overworld.");
+            pbf_press_button(context, BUTTON_MINUS, 20, 105); // open pokedex
```
pifopi, line 414: you are already doing this if you detect overworldin the next detector. Is there any reason not to just remove those 2 lines ?

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -398,6 +399,120 @@ void leave_box_system_to_overworld(const ProgramInfo& info, ConsoleHandle& conso
...
+            );
+        }
+    }
+}
+
+
+void open_recently_battled_from_pokedex(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context){
+    WallClock start = current_time();
+    while (true){
+        if (current_time() - start > std::chrono::minutes(1)){
```
pifopi, line 467: Could that time be reduced? 1 minute sounds huge

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -398,6 +399,120 @@ void leave_box_system_to_overworld(const ProgramInfo& info, ConsoleHandle& conso
...
+                true
+            );
+        }
+    }
+}
+
+
+void open_recently_battled_from_pokedex(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context){
+    WallClock start = current_time();
+    while (true){
```
pifopi, line 466: I think this while is executed once only (then it's stuck in the other while until it return)

pifopi (commented):
I'll continue an other day as it's late and review is quite big

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
NympheaR: Fixed

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
NympheaR: Reduced to 30 seconds

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
NympheaR: Moved time check inside the same while loop

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
NympheaR: Reduced to 30 seconds

SerialPrograms/Source/PokemonSV/Inference/Pokedex/PokemonSV_PokedexMenuDetector.cpp
```diff
@@ -0,0 +1,198 @@
...
+//using std::endl;
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonSV{
+
+using namespace Kernels::Waterfill;
+
+
+const ImageMatch::ExactImageMatcher& RECENTLY_BATTLED_CROPPED(){
```
Mysticial, line 26: Looks like a lot of copy-pasted stuff from the platform bot that isn't needed here?

SerialPrograms/Source/PokemonSV/Inference/Pokedex/PokemonSV_PokedexMenuDetector.h
```diff
@@ -0,0 +1,63 @@
...
+class PokedexMenuWatcher : public DetectorToFinder<PokedexMenuDetector>{
+public:
+    PokedexMenuWatcher(
+        Logger& logger,
+        Color color, bool trigger_if_detected,
+        std::function<void()> on_kill_callback = nullptr,
+        std::chrono::milliseconds duration = std::chrono::milliseconds(250)
+    );
+
+    WallClock last_kill() const{
```
Mysticial, line 43: I don't think this is needed here? Same with all the fields in this class.

SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraRoutines.cpp
```diff
@@ -651,6 +651,35 @@ TeraResult run_tera_summary(
 }
 
 
+void run_from_tera_battle(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context){
```
Mysticial, line 654: This requires having your cursor over the run option right? Based on the naming of this function, it's probably better to move the logic to navigate to the start point into here.

NympheaR: Shifted menu detection and cursor movement under this function as well

SerialPrograms/Source/PokemonSV/Inference/Pokedex/PokemonSV_PokedexMenuDetector.h
NympheaR: Removed leftover platform let's go detection related code

SerialPrograms/Source/PokemonSV/Inference/Pokedex/PokemonSV_PokedexMenuDetector.cpp
NympheaR: Cleaned up so only detection related code is left

SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraRoller.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+        env.console.log("Checking if tera raid is shiny...");
+        env.console.overlay().add_log("Checking Pokédex...", COLOR_WHITE);
+        open_pokedex_from_overworld(env.program_info(), env.console, context);
+        open_recently_battled_from_pokedex(env.program_info(), env.console, context);
+
+        pbf_wait(context, 200);
+
+        // Since encountering the same species within 5 encounters is possible,
+        // loop through all 5 candidates of recently battled pokemon for shinies
```
Mysticial, line 228: So if you see the same species, they get grouped together and not bumped to the top of the list?

NympheaR: Yes if you encounter something that's already within the 5 Recently Battled Pokémon, the order of the list will not change but the Pokédex page will reflect the more recent encounter.

An example to outline this behavior:
Current Pokédex order: Orthworm, Pawmot, Donphan-Male, Ditto, Noivern
After a new encounter with Donphan-Female: Orthworm, Pawmot, Donphan-Female, Ditto, Noivern

Maybe this can be an option for users to pick if they only want to check only the first in the list to save time or be safer and check all five?

NympheaR: This behavior can be useful on single species hunts once the species filter gets added, since you can control it so the species being hunted will always be the first in the list.

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -398,6 +399,112 @@ void leave_box_system_to_overworld(const ProgramInfo& info, ConsoleHandle& conso
...
+void leave_phone_to_overworld(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context){
+    console.log("Exiting to overworld...");
+    pbf_press_button(context, BUTTON_Y, 20, 250);
+
+    WallClock start = current_time();
+    while (true){
+        if (current_time() - start > std::chrono::seconds(30)){
+            throw OperationFailedException(
+                ErrorReport::SEND_ERROR_REPORT, console,
+                "leave_phone_to_overworld(): Failed to open return to overworld after 30 seconds.",
```
Mysticial, line 485: > Failed to open return to overworld after 30 seconds.

PR #350, opened 2023-08-08
Clothing buyer

kichithewolf:
Clothing buyer for SV. Supports shops with both single and multiple categories of clothing.

also change default and clarify in-game money limit for tournament farmer.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_ClothingBuyer.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_ClothingBuyer.cpp
@@ -77,28 +77,32 @@ void ClothingBuyer::program(SingleSwitchProgramEnvironment& env, BotBaseContext&
     */
 
     uint8_t category_rotation_count = 0;
+    bool finish_program = false;
     while (category_rotation_count < NUM_CATEGORY) {
         pbf_press_button(context, BUTTON_A, 20, 100);
 
         AdvanceDialogWatcher already_purchased(COLOR_RED);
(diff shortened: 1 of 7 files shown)


SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_ClothingBuyer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    * Stop conditions: Out of money/lp, hit catogory count
+    */
+
+    uint8_t category_rotation_count = 0;
+    while (category_rotation_count < NUM_CATEGORY) {
+        pbf_press_button(context, BUTTON_A, 20, 100);
+
+        AdvanceDialogWatcher already_purchased(COLOR_RED);
+        PromptDialogWatcher buy_yes_no(COLOR_CYAN);
+        PromptDialogWatcher wear_yes_no(COLOR_CYAN);
```
Mysticial, line 85: Do we need both of these? They look the same.

SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_ClothingBuyer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        AdvanceDialogWatcher afford_yes_no(COLOR_RED);
+
+        int ret = wait_until(
+            env.console, context,
+            std::chrono::seconds(10),
+            { already_purchased, buy_yes_no }
+        );
+        if (ret == 0) {
```
Mysticial, line 93: Use a switch statement instead of chaining ifs.

PR #352, opened 2023-08-14
Add Mystery Gift mode to SV FCE

NympheaR:


SerialPrograms/Source/PokemonSV/Programs/FastCodeEntry/PokemonSV_CodeEntry.cpp
```diff
@@ -72,19 +72,22 @@ const char* normalize_code(std::string& normalized_code, const std::string& code
 
     switch (normalized_code.size()){
```
pifopi, line 73: couldn't you return here if override_mode ?

SerialPrograms/Source/PokemonSV/Programs/FastCodeEntry/PokemonSV_FastCodeEntry.cpp
```diff
@@ -125,6 +126,14 @@ void FastCodeEntry::program(MultiSwitchProgramEnvironment& env, CancellableScope
         return;
     }
 
+    if (MODE == Mode::MYSTERY_GIFT){
+        const char* error = enter_code(env, scope, settings, CODE, true, true);
+        if (error){
```
pifopi, line 131: I don't think this if can ever happen as you always return an empty string

SerialPrograms/Source/PokemonSV/Programs/FastCodeEntry/PokemonSV_CodeEntry.cpp
NympheaR: added return if override before the other checks

SerialPrograms/Source/PokemonSV/Programs/FastCodeEntry/PokemonSV_FastCodeEntry.cpp
NympheaR: merged the logic with the normal mode check and have error report only if on normal mode

pifopi (approved):

PR #357, opened 2023-08-25
Add event/non-event raid filter option to Tera Roller

NympheaR:


SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraRoller.cpp
```diff
@@ -114,6 +114,16 @@ TeraRoller::TeraRoller()
           LockWhileRunning::UNLOCKED,
           false
     )
+    , EVENT_CHECK_MODE(
+        "<b>Event Tera Raid Action:</b><br>Choose how to interact with event/non-event raids.",
+        {
+            {EventCheckMode::CHECK_ALL,             "check_all",        "Check all raids / No ongoing tera raid event"},
```
pifopi, line 120: Why specify no ongoing tera raid event ? Would the "only non event" break in that case ?

pifopi (approved):

SerialPrograms/Source/PokemonSV/Programs/TeraRaids/PokemonSV_TeraRoller.cpp
NympheaR: Updated the wording for more clarity on which option the user should use

PR #364, opened 2023-09-14
Bloodmoon ursaluna stats reset

kichithewolf:
Spams first attack against Ursaluna. Will switch out if Pokemon faints. I used Walking Wake and spammed Hydro Stream for a two turn kill.
Stand in front of Perrin. Have 5 Pokemon in your party.


Gin890:
Great!

PR #369, opened 2023-09-24
stats reset: turn 1 quickball, move table options

kichithewolf:
catch rate improvements

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
@@ -162,6 +162,40 @@ void StatsReset::enter_battle(SingleSwitchProgramEnvironment& env, BotBaseContex
     }
 }
 
+void StatsReset::open_ball_menu(SingleSwitchProgramEnvironment& env, BotBaseContext& context) {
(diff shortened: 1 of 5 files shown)


SerialPrograms/Source/PokemonSV/Programs/General/PokemonSV_StatsReset.cpp
```diff
@@ -297,11 +342,37 @@ bool StatsReset::run_battle(SingleSwitchProgramEnvironment& env, BotBaseContext&
                     }
                 }
                 else {
-                    //Navigate to correct ball and repeatedly throw it until caught
-                    pbf_press_button(context, BUTTON_X, 20, 100);
-                    context.wait_for_all_requests();
-
                     BattleBallReader reader(env.console, LANGUAGE);
+                    std::string ball_reader = "";
```
Mysticial, line 346: Probably a good idea to wrap this in a function to avoid duplicating the code above.

This seems to suggest that navigating into the ball selection logic should belong in the `move_to_ball()` function or at least somewhere up there since I've seen the dropped A press break other programs. That will be a cleanup that I'll do at a later time.

kichithewolf: Done.

PR #380, opened 2023-11-28
Upgrade qt

pifopi:


pifopi:
qt 6.6.2 isn't released yet. Once it will, it should have this fix https://bugreports.qt.io/browse/QTBUG-117765

PR #383, opened 2023-12-19
dpp integration for macos/arm64

naussika:
- add dpp v10.0.22 static lib;
	- enable PA_DPP flag;

### DIFF
--- a/SerialPrograms/Source/Integrations/DiscordIntegrationSettings.cpp
+++ b/SerialPrograms/Source/Integrations/DiscordIntegrationSettings.cpp
@@ -123,8 +123,8 @@ void DiscordIntegrationSettingsOption::value_changed(){
 #if (defined PA_SLEEPY || defined PA_DPP)
     bool options_enabled = this->enabled();
     switch (library0){
-#ifdef PA_SLEEPY
     case Library::SleepyDiscord:{
+#ifdef PA_SLEEPY
         options_enabled &= !SleepyDiscordRunner::is_running();
         ConfigOptionState state = options_enabled ? ConfigOptionState::ENABLED : ConfigOptionState::DISABLED;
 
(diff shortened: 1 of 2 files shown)


naussika:
note that in theory i could build a universal static library (both amr64 and x86_64) for `libdpp.a`, however the Xcode on my machine doesn't support x86_64 anymore (the previous version could tho).

enabling dpp on `macos/x86_64` would be trivial tho, which just requires a x86_64 static `libdpp.a` and some minor changes in `CMakeFiles.txt`.

btw `libdpp_macos_arm64.a` was built on macOS 14.2.

for CI test building maybe a new job with macos-14 needs to be added.

SerialPrograms/Source/Integrations/DiscordIntegrationSettings.cpp
```diff
@@ -204,16 +204,16 @@ DiscordIntegrationSettingsWidget::DiscordIntegrationSettingsWidget(QWidget& pare
         button_start, &QPushButton::clicked,
         this, [&value](bool) {
             switch (value.library0){
-#ifdef PA_SLEEPY
             case DiscordIntegrationSettingsOption::Library::SleepyDiscord:
+#ifdef PA_SLEEPY
```
Mysticial, line 208: Why are you moving these if-defs? These Sleepy cases are not supposed to be compiled when it's disabled.

naussika: because otherwise compilation failed due to `-Werror`

```
[  1%] Building CXX object CMakeFiles/SerialPrograms.dir/Source/Integrations/DiscordIntegrationSettings.cpp.o
build/Arduino-Source/SerialPrograms/Source/Integrations/DiscordIntegrationSettings.cpp:125:13: error: enumeration value 'SleepyDiscord' not handled in switch [-Werror,-Wswitch]
    switch (library0){
            ^~~~~~~~
build/Arduino-Source/SerialPrograms/Source/Integrations/DiscordIntegrationSettings.cpp:125:13: note: add missing switch cases
    switch (library0){
            ^
1 error generated.
```

Mysticial: You can use a `default;` to swallow everything else.

PR #388, opened 2023-12-24
Update Flying Trail Path, Add Fail Detection

NympheaR:


Mysticial:
Why do you need a separate BlackDialogBoxDetector. Is it different from the existing DialogBoxDetector that will detect either black or white?

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_FlyingTrialFarmer.cpp
```diff
@@ -66,10 +74,42 @@ FlyingTrailFarmer::FlyingTrailFarmer()
...
+        switch (ret_finish){
+        case 0: // dialog
+            trial_failed = false;
+            continue;
+        case 1: // overworld
+            return trial_failed;
+        default:
+            throw OperationFailedException(
+                ErrorReport::SEND_ERROR_REPORT, env.console,
+                "No recognized state after 10000 B presses.",
```
Mysticial, line 102: I think this 10000 presses is wrong. lol Should be "after 80 seconds".

PR #401, opened 2024-01-20
Increase B presses until overworld

NympheaR:


pifopi:
Did you forget the implementation of that new method ? I don't see it

pifopi (approved):

PR #405, opened 2024-03-03
Update build guide for Qt6.3.2

jw098:
In particular, I removed bullet 9, which mentions editing CMakeLists.txt. I feel this was confusing since QT_MAJOR was already set to 6.


Mysticial:
Let me know when this is ready since I got a lot of notifs from the last one and I can't tell if it's will WIP.

jw098:
Sorry about that. I'm new to using git and I realized that my last pull request was working off the main branch instead of a separate branch.

I'm basically done. The only other change I would suggest is adding another screenshot for step 7, but the directory for BuildSetup is excluded from git.

Windows-desktop-Configuration

Mysticial:
You should be able to override it. Probably depends on what git client you're using.

pifopi:
git add --force path/to/your/file.txt
For the command if you are using command line

jw098:
Thanks. I didn't know overriding was possible.

I'm now done with my changes.

PR #407, opened 2024-03-03
BBQ Solo Farmer

kichithewolf:
BBQ json: https://github.com/PokemonAutomation/Packages/pull/22
Setup/Instructions: https://github.com/PokemonAutomation/ComputerControl/pull/177

All singleplayer quests except "Pickup 10 items" are automated. That quest will be skipped, no rerolling, it only takes up one slot. No intention of doing multiplayer at this time, I've never found a ditto block in my life.

Some quests more reliable than others, ex. hatch egg and make sandwich are solid but the catch/photo quests are more sensitive to weather-induced lag since they have to navigate the area. I tried aiming for locations that are less likely to lag but there's no avoiding it outright. "Give your Pokemon a bath" depends on what Pokemon it ends up washing, it just moves the brush in a + shape. The program will retry quests anyway, so it will eventually complete the quests. I was able to do a successful run of 100 quests a few times.

Shiny checking - there's checking for overworld encounters that should work in theory, in practice all the targets are fixed spawns so they're full odds and I have no way of really testing it.

There are still improvements that can be done, quest processing logic was written with the assumption that most quests could not be automated so could be tweaked a bit, travel 500 meters could use sound detection (I don't have the sound file for it) to check when it is complete, could probably keep track of egg/TM position, etc. This things is 10 different programs in a trenchcoat, take your time reviewing. I need a break since it's been 3 weeks.

PR #409, opened 2024-03-24
sv terarium navigation

kichithewolf:
pulled out from https://github.com/PokemonAutomation/Arduino-Source/pull/407

commented out jump_glide_fly() since that requires the other BBQ files.
new escape_stuck_on_wall() added to SV_Navigation.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.h
+++ b/SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.h
@@ -67,6 +67,10 @@ void leave_phone_to_overworld(const ProgramInfo& info, ConsoleHandle& console, B
 //  by other map icons on the most zoomed-in level of the map.
 void fly_to_closest_pokecenter_on_map(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context);
 
+// Attempt to escape being stuck on a wall.
+// Repeatedly center camera and try to backwards jump off.
+void escape_stuck_on_wall(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context);
+
 
 }
 }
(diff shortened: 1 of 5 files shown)


SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.h
```diff
@@ -67,6 +67,10 @@ void leave_phone_to_overworld(const ProgramInfo& info, ConsoleHandle& console, B
 //  by other map icons on the most zoomed-in level of the map.
 void fly_to_closest_pokecenter_on_map(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context);
 
+// Attempt to escape being stuck on a wall.
+// Repeatedly center camera and try to backwards jump off.
```
Mysticial, line 71: Since the end state here is the map being open, we should include that in the function name somehow.

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Terarium.cpp
```diff
@@ -0,0 +1,371 @@
...
+                    "Unable to fly to Chargestone Cavern!",
+                    true
+                );
+            }
+        }
+    }
+}
+
+/*
+void jump_glide_fly(ConsoleHandle& console, BotBaseContext& context, BBQOption& BBQ_OPTIONS, uint16_t hold_up, uint16_t flight_wait, uint16_t drop_time) {
```
Mysticial, line 342: I think we can enable this here. The only option you're using is the inverted flight which is just a boolean. So pass a boolean instead of the whole optino.

SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Terarium.cpp
```diff
@@ -0,0 +1,371 @@
...
+                        true
+                    );
+                }
+            }
+        }
+    }
+    context.wait_for_all_requests();
+}
+
+void central_to_polar_rest(const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context) {
```
Mysticial, line 125: A lot of code duplication here with all the different fly locations.

Since they differ only by the joystick move and the exception string, you move everything into a helper function that takes just the joystick command and the location name for the exception string.

PR #411, opened 2024-03-25
separate sandwich maker functions

kichithewolf:
The old run_sandwich_maker() is replaced with make_sandwich_option(). This uses the ingredients/presets from the sandwich options. It processes the option and then calls make_sandwich_preset().

To make a custom preset sandwich:
fillings = {{"apple", (uint8_t)1}};
condiments = {{"marmalade", (uint8_t)1}};
make_sandwich_preset(env, context, language, fillings, condiments);

Removed build_great_peanut_butter_sandwich and used the custom sandwich builder in its place. The replacement is a bit messy because it skips the custom sandwich menu and goes straight to the new run_sandwich_maker(). I didn't want to use make_sandwich_preset() since selecting the recipe is faster than selecting custom ingredients.

Tested that they still work but otherwise didn't touch the two Herba lv3 egg sandwiches.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.cpp
@@ -253,41 +253,44 @@ void picnic_at_zero_gate(const ProgramInfo& info, ConsoleHandle& console, BotBas
     picnic_from_overworld(info, console, context);
 }
 
-bool eat_egg_sandwich_at_picnic(const ProgramInfo& info, AsyncDispatcher& dispatcher, ConsoleHandle& console, BotBaseContext& context,
+bool eat_egg_sandwich_at_picnic(SingleSwitchProgramEnvironment& env, AsyncDispatcher& dispatcher, ConsoleHandle& console, BotBaseContext& context,
     EggSandwichType sandwich_type, Language language)
 {
     // Move forward to table to make sandwich
     pbf_move_left_joystick(context, 128, 0, 30, 40);
     context.wait_for_all_requests();
 
-    clear_mons_in_front(info, console, context);
-    if (enter_sandwich_recipe_list(info, console, context) == false){
+    clear_mons_in_front(env.program_info(), console, context);
+    if (enter_sandwich_recipe_list(env.program_info(), console, context) == false){
         return false;
     }
     switch (sandwich_type){
     case EggSandwichType::GREAT_PEANUT_BUTTER:
     {
-        if (select_sandwich_recipe(info, console, context, 17) == false){
+        if (select_sandwich_recipe(env.program_info(), console, context, 17) == false){
             // cannot find the sandwich recipe, either user has not unlocked it or does not have enough ingredients:
             return false;
         }
-        build_great_peanut_butter_sandwich(info, dispatcher, console, context);
+        std::map<std::string, uint8_t> fillings = { {"banana", (uint8_t)1} };
+        std::vector<std::string> fillings_sorted;
+        fillings_sorted.push_back("banana");
+        int plates = 1;
+        run_sandwich_maker(env, context, language, fillings, fillings_sorted, plates);
         break;
     }
     case EggSandwichType::TWO_SWEET_HERBS:
     case EggSandwichType::SALTY_SWEET_HERBS:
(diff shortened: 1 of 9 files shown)


SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.h
```diff
@@ -45,7 +45,7 @@ void picnic_at_zero_gate(const ProgramInfo& info, ConsoleHandle& console, BotBas
 // - Two-sweet-herbs, bitter-sweet-herbs or salty-sweet-herbs custom sandwich to gain egg power Lv 3, must have enough ingredinets and
 //   provide game language for OCR ingredient lists.
 // Return false if no needed sandwich ingredients or recipe.
-bool eat_egg_sandwich_at_picnic(const ProgramInfo& info, AsyncDispatcher& dispatcher, ConsoleHandle& console,
+bool eat_egg_sandwich_at_picnic(SingleSwitchProgramEnvironment& env, AsyncDispatcher& dispatcher, ConsoleHandle& console,
```
Mysticial, line 48: I'm a bit iffy about this change since it means we can't call this function in a multi-Switch program. But at the same time, we currently don't have any multi-Switch programs that make sandwiches.

Looking further down, it looks like the original sandwich maker already took `SingleSwitchProgramEnvironment` as a parameter. Tracing it further down, the only part that it uses is log and the overlay - neither of which are special to `SingleSwitchProgramEnvironment`.

The "proper" thing to do here (which should've been done in the beginning a long time ago), is to replace `SingleSwitchProgramEnvironment& env` with `const ProgramInfo& info` and `ConsoleHandle& console`. Since `console` is already passed, it looks like the best thing here is to leave it as `const ProgramInfo& info`. Then all logging calls `env.log(...)` become `console.log(...)` and all the overlay calls `env.console.overlay()` become just `console.overlay()`.

SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_SandwichMaker.cpp
```diff
@@ -53,7 +53,7 @@ SandwichMaker::SandwichMaker()
 void SandwichMaker::program(SingleSwitchProgramEnvironment& env, BotBaseContext& context) {
     assert_16_9_720p_min(env.logger(), env.console);
 
-    run_sandwich_maker(env, context, SANDWICH_OPTIONS);
+    make_sandwich_option(env.program_info(), env.realtime_dispatcher(), env.console, context, SANDWICH_OPTIONS);
```
Mysticial, line 56: Looks like I overlooked the dispatcher.

Then let's pass `ProgramEnvironment` instead. It's the parent class of Single/MultiSwitchProgramEnvironment and has both the `program_info()` and the dispatchers (as well as the logger).

SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.cpp
```diff
@@ -253,41 +253,44 @@ void picnic_at_zero_gate(const ProgramInfo& info, ConsoleHandle& console, BotBas
     picnic_from_overworld(info, console, context);
 }
 
-bool eat_egg_sandwich_at_picnic(const ProgramInfo& info, AsyncDispatcher& dispatcher, ConsoleHandle& console, BotBaseContext& context,
+bool eat_egg_sandwich_at_picnic(SingleSwitchProgramEnvironment& env, AsyncDispatcher& dispatcher, ConsoleHandle& console, BotBaseContext& context,
```
Mysticial, line 256: `SingleSwitchProgramEnvironment` -> `ProgramEnvironment`

Mysticial: And also get rid of the `AsyncDispatcher& dispatcher` since that's inside `env`.

SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggRoutines.h
Mysticial: So same as above:

`SingleSwitchProgramEnvironment -> ProgramEnvironment`

and remove `AsyncDispatcher& dispatcher` from parameter since it's inside `env`.

PR #413, opened 2024-03-30
Fixes compilation on macOS

naussika:
- adding missing #include headers

### DIFF
--- a/SerialPrograms/Source/Integrations/DppIntegration/DppUtility.h
+++ b/SerialPrograms/Source/Integrations/DppIntegration/DppUtility.h
@@ -3,6 +3,7 @@
 #define DPP_UTILITY_H
 
 #include <dpp/dpp.h>
+#include <atomic>
 #include "Common/Cpp/AbstractLogger.h"
 
 namespace PokemonAutomation{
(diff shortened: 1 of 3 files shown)


SerialPrograms/CMakeLists.txt
```diff
@@ -2082,7 +2082,7 @@ else()
     # Add -Wno-c11-extensions to avoid clang gives
     # /usr/local/Cellar/opencv/4.5.5_3/include/opencv4/opencv2/core/mat.inl.hpp:2116:9: error: '_Atomic' is a C11 extension
     # when compiling OpenCV
-    target_compile_options(SerialPrograms PRIVATE -Wall -Wextra -Wpedantic -Werror -Wno-c11-extensions)
+    target_compile_options(SerialPrograms PRIVATE -O2 -Wall -Wextra -Wpedantic -Werror -Wno-c11-extensions)
```
Mysticial, line 2085: Really? Were we really missing this the entire time?

SerialPrograms/Source/Integrations/DppIntegration/DppUtility.h
```diff
@@ -3,6 +3,7 @@
 #define DPP_UTILITY_H
 
 #include <dpp/dpp.h>
+#include <atomic>
```
Mysticial, line 6: Why is this needed here? Nothing in this file uses <atomic>.

PR #414, opened 2024-03-31
Tournament Farmer 2

jw098:
- Added new FastTravelDetector (along with tests), to detect win vs loss instead of using timeout.
- added more print statements for debugging.
- adjusted the stick movement for `return_to_academy_after_loss`, to make it more consistently move to the fly point. Also, gives it multiple attempts if fails to fly there. To do this, `fly_to_overworld_from_map` now uses `MapFlyMenuWatcher` and `MapDestinationMenuWatcher`, since `check_fly_menuitem` is set to true.

### DIFF
--- a/SerialPrograms/Source/CommonFramework/ImageMatch/WaterfillTemplateMatcher.cpp
+++ b/SerialPrograms/Source/CommonFramework/ImageMatch/WaterfillTemplateMatcher.cpp
@@ -74,21 +74,18 @@ bool WaterfillTemplateMatcher::check_aspect_ratio(size_t candidate_width, size_t
     
     bool pass = m_aspect_ratio_lower <= error && error <= m_aspect_ratio_upper;
 
-    if (!pass && PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING){
-        cout << "Failed to pass WaterfillTemplateMatcher aspect ratio check (W/H): "
-             << "expected: " << image_template.width() << " : " << image_template.height()
-             << " (" << (double)image_template.width()/image_template.height() << "), "
-             << "actual: " << candidate_width << " : " << candidate_height 
-             << " (" << (double)candidate_width/candidate_height << ") "
-             << "error: " << error << " bound [" << m_aspect_ratio_lower << ", " << m_aspect_ratio_upper << "]" << endl;
-    }
-    else if (pass && PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING){
-        cout << "Passed WaterfillTemplateMatcher aspect ratio check (W/H): "
-                << "expected: " << image_template.width() << " : " << image_template.height()
-                << " (" << (double)image_template.width()/image_template.height() << "), "
-                << "actual: " << candidate_width << " : " << candidate_height 
-                << " (" << (double)candidate_width/candidate_height << ") "
-                << "error: " << error << " bound [" << m_aspect_ratio_lower << ", " << m_aspect_ratio_upper << "]" << endl;
+    if (PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING){
+        if (!pass){
+            cout << "Failed to pass WaterfillTemplateMatcher aspect ratio check (W/H): ";
+        }
+        else {
+            cout << "Passed WaterfillTemplateMatcher aspect ratio check (W/H): ";
+        }
+        cout << "expected: " << image_template.width() << " : " << image_template.height()
+            << " (" << (double)image_template.width()/image_template.height() << "), "
+            << "actual: " << candidate_width << " : " << candidate_height 
+            << " (" << (double)candidate_width/candidate_height << ") "
+            << "error: " << error << " bound [" << m_aspect_ratio_lower << ", " << m_aspect_ratio_upper << "]" << endl;
     }
 
     return pass;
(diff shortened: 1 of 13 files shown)


SerialPrograms/Source/CommonFramework/ImageMatch/WaterfillTemplateMatcher.cpp
```diff
@@ -82,6 +82,15 @@ bool WaterfillTemplateMatcher::check_aspect_ratio(size_t candidate_width, size_t
              << " (" << (double)candidate_width/candidate_height << ") "
              << "error: " << error << " bound [" << m_aspect_ratio_lower << ", " << m_aspect_ratio_upper << "]" << endl;
     }
+    else if (pass && PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING){
```
Mysticial, line 85: Since both halves of this if-statement require `PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING`. Might as well just wrap the entire thing with an if-statement for that, then branch it into pass = true or false.

SerialPrograms/Source/CommonFramework/ImageMatch/WaterfillTemplateMatcher.cpp
```diff
@@ -95,6 +104,12 @@ bool WaterfillTemplateMatcher::check_area_ratio(double candidate_area_ratio) con
              << "Expected: " << m_area_ratio << ", actual: " << candidate_area_ratio << ", "
              << "error: " << error << " bound [" << m_area_ratio_lower << ", " << m_area_ratio_upper << "]" << endl;
     }
+    else if (pass && PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING){
```
Mysticial, line 107: Same thing here.

SerialPrograms/Source/CommonFramework/ImageMatch/WaterfillTemplateMatcher.cpp
jw098: Fixed. I also did more refactoring of the debugging print statements.

SerialPrograms/Source/CommonFramework/ImageMatch/WaterfillTemplateMatcher.cpp
```diff
@@ -140,13 +131,50 @@ double WaterfillTemplateMatcher::rmsd_original(const ImageViewRGB32& original_im
...
 
     double rmsd = this->rmsd(extract_box_reference(original_image, object));
-    if (PreloadSettings::debug().IMAGE_TEMPLATE_MATCHING){
-        cout << "Passed aspect and area ratio check, rmsd = " << rmsd << endl;
-    }
-   
 
     return rmsd;
 }
+void WaterfillTemplateMatcher::check_aspect_ratio_debug(size_t candidate_width, size_t candidate_height) const {
```
Mysticial, line 137: Not a real fan of duplicating the logic between the main code and the debugging code here. Because it creates an opportunity to for them to accidentally get out-of-sync which would make the debug versions behave differently from the main ones.

SerialPrograms/Source/PokemonSV/Inference/Map/PokemonSV_FastTravelDetector.cpp
```diff
@@ -0,0 +1,128 @@
...
+    const std::vector<std::pair<uint32_t, uint32_t>> filters = {
+        {combine_rgb(0, 0, 80), combine_rgb(140, 140, 255)},
+        {combine_rgb(5, 100, 200), combine_rgb(60, 135, 255)},
+    };
+
+    const double min_object_size = 400.0;
+    // need to keep RMSD threshold below 100, else will false positive the blue dots at blueberry academy
+    const double rmsd_threshold = 80.0;
+
+    const double screen_rel_size = (screen.height() / 1080.0);
```
Mysticial, line 70: Has this been tested at 720p?

jw098: Good point. I didn't. It does fail under certain circumstances at 720p (mostly at Blueberry academy). I'll try to see if I can fix it.

jw098: I tried adjusting my detector to also work at 720p at the blueberry academy. With my current technique, this isn't possible without also increasing false positives. It works approximately 75-80% of the time depending on map rotation.

Given that the detector is able to do its job for the Ace Tournament, and that it works well everywhere else, I suggest we just leave it alone for now. Especially since nothing else uses this detector for now. If we do need better detection at the Blueberry academy in the future, I do have some ideas, such as using a two step waterfill+template match technique. Also, I wrote out the detector's limitations and ideas for future improvement as a comment in the header file.

Mysticial: What do the false positives look like?

jw098: Turquoise colored solid squares. These have a good aspect ratio and area ratio and their rmsd is 80. We can't lower the rmsd threshold because then you false negative at 720p, even for perfect candidates found by the waterfill.

I believe this is because to select for the fly target at 720p at blueberry academy, you need very narrow filters that include your target but not the rest of the blue background. The fly target at 720p has turquoise pixels. 

We only have an issue with getting false positive turquoise squares with narrow filters. With broader filters, it would select more of its surrounding pixels, causing it to fail  the rmsd,  aspect/area ratio checks as expected.

Mysticial: If you're talking about the Turquoise square in the center area of the Blueberry map, that looks nothing like the fly icon. So it feels like something else is wrong. I'd try doing a side-by-side comparison of the cropped object along with the actual fly icon.

jw098: I pushed a new change. It now detects the white, winged portion of the fast travel icon. It seems to work much better. It works even at the blueberry academy at 720p, and doesn't false positive against the winged icon at Pokemon centers (sadly, it still doesn't work indoors at the blueberry academy, due to obstruction from the radar dot overlapping with the fast travel icon. Although, this could be fixed with a custom fast travel icon template.). I was able to get it to work with a single filter, without much need for fine tuning, so I'm hoping it will also be robust against different capture card settings.

PR #416, opened 2024-04-01
bbq quest detector and reader

kichithewolf:
Split out from https://github.com/PokemonAutomation/Arduino-Source/pull/407.


SerialPrograms/Source/PokemonSV/Inference/PokemonSV_BlueberryQuestDetector.cpp
```diff
@@ -0,0 +1,113 @@
...
+            results.emplace(result.first, result.second);
+        }
+            
+    }
+
+    if (results.empty()){
+        return "";
+    }
+    
+    return results.begin()->second.token;
```
Mysticial, line 78: Should we do something about ambiguous matches or multiple matches?

PR #417, opened 2024-04-06
Scatterbug: Add Notifications

NympheaR:


SerialPrograms/Source/PokemonSV/Programs/ShinyHunting/PokemonSV_ShinyHunt-Scatterbug.cpp
```diff
@@ -270,6 +274,7 @@ void ShinyHuntScatterbug::handle_battles_and_back_to_pokecenter(SingleSwitchProg
                 }
             }catch (ProgramFinishedException&){
                 GO_HOME_WHEN_DONE.run_end_of_program(context);
+                send_program_finished_notification(env, NOTIFICATION_PROGRAM_FINISH);
```
Mysticial, line 277: I believe this is sent automatically if it's thrown as an exception.

NympheaR: updated

SerialPrograms/Source/PokemonSV/Programs/ShinyHunting/PokemonSV_ShinyHunt-Scatterbug.cpp
```diff
@@ -204,7 +204,6 @@ void ShinyHuntScatterbug::program(SingleSwitchProgramEnvironment& env, BotBaseCo
             send_program_status_notification(env, NOTIFICATION_STATUS_UPDATE);
         }catch(ProgramFinishedException&){
             GO_HOME_WHEN_DONE.run_end_of_program(context);
-            send_program_finished_notification(env, NOTIFICATION_PROGRAM_FINISH);
```
Mysticial: So this one isn't actually redundant since it returns instead of throws. So the inconsistency here with the one below is that one throws while the other returns.

Probably better to make them both rethrow the exception.

PR #420, opened 2024-04-15
Material farmer

jw098:
- Added material farmer. Farms for Happiny dust for the item printer. Based on the Scatterbug program. But does not detect shinies.
- One weakness is that due to the terrain, the character can't walk in a straight line. For some reason, it likes to walk up hills. So, it's often walking up against the walls. Due to variation in the way the game takes inputs, you can sometimes be walking up against the left wall and sometimes the right wall. As far as I can see, it can't easily be fixed without a way to detect player direction and adjust accordingly. Still, it earns approximately 150-300 Happiny dust per sandwich. So, you'll easily max out in 2-4 hours.
- As I mentioned, this program is based on the Scatterbug program. There is some duplicated code. Particularly `handle_battles_and_back_to_pokecenter`. I'm not sure if it's better to refactor it so both programs can use the same function. Or if it's better to leave them separate, in case the programs diverge more in the future.


Mysticial:
Any idea how much faster is this vs. running platform bot with no sandwiches in terms of total "value"?

jw098:
I haven't tested it. I'll have to get back to you on this. If I remember correctly, platform bot is decent if you use a dragon encounter sandwich. One problem though is that I think platform bot stops on shinies.

jw098:
I did some testing, although with low sample sizes (only 30-60 minutes of testing each). With Chansey farming, I get around 5.6 of item printer value per minute. Area zero with dragon sandwich: 4.6 value/min. Area zero no sandwich: 4 value/min. 

One consideration, however, is that with Chansey farming, you will max out within approximately 2-3 hours. With area zero farming without a sandwich, it takes much longer to max out because the materials are spread out over several pokemon.

I wonder if it would be good to just have 2 different material farmers. So people have the option of faster farming, or slower farming, but has a higher max.

SerialPrograms/Source/PokemonSV/PokemonSV_Panels.cpp
```diff
@@ -108,6 +110,7 @@ std::vector<PanelEntry> PanelListFactory::make_panels() const{
         ret.emplace_back(make_single_switch_program<BBQSoloFarmer_Descriptor, BBQSoloFarmer>());
         ret.emplace_back(make_single_switch_program<ItemPrinterRNG_Descriptor, ItemPrinterRNG>());
     }
+    ret.emplace_back(make_single_switch_program<MaterialFarmer_Descriptor, MaterialFarmer>());
```
Mysticial, line 113: Put all new programs behind dev mode.

SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_SandwichRoutines.cpp
```diff
@@ -637,8 +637,9 @@ void make_two_herbs_sandwich(
 
 void make_sandwich_option(ProgramEnvironment& env, ConsoleHandle& console, BotBaseContext& context, SandwichMakerOption& SANDWICH_OPTIONS){
     const Language language = SANDWICH_OPTIONS.LANGUAGE;
-    if (language == Language::None){
-        throw UserSetupError(console.logger(), "Must set game langauge option to read ingredient lists.");
+
+    if (language == Language::None) {
+        throw UserSetupError(console.logger(), "Must set game language option to read ingredient lists.");
```
Mysticial, line 642: This might've crossed with a global code formatting change yesterday where I removed spaces between ")" and "{"

jw098:
The material farmer should be done now.
- Updated the Let's go path. It can now farm 500 Happiny dust per sandwich.
- fixed bug with handling recoverable errors in both Material Farmer and Scatterbug program.

SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
```diff
@@ -30,6 +30,7 @@ const std::set<std::string> TOKENS{
     "6643d9fe87b3e54dc75dfac8ac22f0cc8bd17f6a8a786debf5fc4c517ee65469",
     "8e48e38e49bffc8462ada9d2d9d850d5b3b5c9529d20978c09bc548bc9a614a4",
     "7694adee4419d62c6a923c4efc9e7b41def7b96bb84ea882701b0bf2e8c13bee",
+    "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08", // test
```
pifopi, line 33: put your name on it so that we know it's coming from you

pifopi (approved):
This MR needs to be updated to latest before we can merge it (there is a conflict)

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmer.cpp
```diff
@@ -0,0 +1,566 @@
...
+    // - Ensure audio input is enabled
+    LetsGoKillSoundDetector audio_detector(env.console, [](float){ return true; });
+    wait_until(
+        env.console, context,
+        std::chrono::milliseconds(1100),
+        {
+            static_cast<AudioInferenceCallback&>(audio_detector)
+        }
+    );
+    audio_detector.throw_if_no_sound(std::chrono::milliseconds(1000));
```
Mysticial, line 172: Does this need to happen explicitly at the beginning here? Or can be done as part of whatever uses sound later?

jw098: It can be done later. But I thought it would be better from a user experience point of view if the user setup error was thrown early, as opposed to 5-10 minutes later, when the user may have stepped away.

Do you feel it's better to do this check later?

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmer.cpp
```diff
@@ -0,0 +1,566 @@
...
+            {battle_menu}
+        );
+        if (ret == 0){
+            env.console.log("Detected battle. Now running away.", COLOR_PURPLE);
+            env.console.overlay().add_log("Detected battle. Now running away.");
+            try{
+                run_from_battle(env.program_info(), env.console, context);
+            }catch (OperationFailedException& e){
+                throw FatalProgramException(std::move(e));
+            }
```
Mysticial, line 557: So if I understand the logic here correctly. If you get attacked, you fly back and try again right? So that way if you encounter a shiny (which you cannot kill), flying away will force despawn it so won't get in the way.

(I'm thinking in some later version after everything goes in, we can re-add the shiny detection + action table. lol)

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmer.cpp
```diff
@@ -0,0 +1,566 @@
...
+            use_lets_go_to_clear_in_front(console, context, *m_encounter_tracker, throw_ball_if_bubble, [&](BotBaseContext& context){
+                // Do the following movement while the Let's Go pokemon clearing wild pokemon.
+                env.console.log("Move-forward iteration number: " + std::to_string(i + 1) + "/" + std::to_string(total_iterations), COLOR_PURPLE);
+
+                lets_go_movement1(context);
+            });
+        }
+    };
+
+    move_forward_with_lets_go(total_iterations);
```
Mysticial, line 501: Why split this into lambda instead of just inlining the code?

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmer.cpp
jw098: Yes. That's the logic.

If you want shiny detection, ShinyHuntScatterbug::handle_battles_and_back_to_pokecenter already does this. It's basically the same function, except I just run away by default in my version, whereas the scatterbug version uses the action table.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmer.cpp
jw098: Now that you mention it, I could. It's like this because I based this off of the Scatterbug program, where a lambda is needed.

PR #421, opened 2024-04-21
Another fix to close game sequence

jw098:
There's another report of the zoom in issue. 
https://discord.com/channels/695809740428673034/711649658220314635/1231603132509720766
They report it only occurs every 150 resets, so I'm guessing that sometimes the program isn't respecting the delay between the 2 home button presses. Because from my testing, the zoom feature should only activate if the delay between home button presses is less than 40 ticks (currently, the delay is 100 ticks). I'm hoping that increasing the delay between presses and using `wait_for_all_requests` will solve the issue.

### DIFF
--- a/SerialPrograms/Source/NintendoSwitch/Commands/NintendoSwitch_Commands_Routines.cpp
+++ b/SerialPrograms/Source/NintendoSwitch/Commands/NintendoSwitch_Commands_Routines.cpp
@@ -6,6 +6,9 @@
 
 #include <sstream>
 #include "ClientSource/Libraries/MessageConverter.h"
+#include "NintendoSwitch/Inference/NintendoSwitch_DetectHome.h"
+#include "CommonFramework/VideoPipeline/VideoFeed.h"
+#include "CommonFramework/InferenceInfra/InferenceRoutines.h"
 #include "NintendoSwitch/Commands/NintendoSwitch_Commands_Superscalar.h"
 #include "NintendoSwitch_Commands_Routines.h"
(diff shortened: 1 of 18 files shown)


SerialPrograms/Source/NintendoSwitch/Commands/NintendoSwitch_Commands_Routines.cpp
```diff
@@ -32,7 +32,8 @@ void close_game(BotBaseContext& context){
     pbf_press_dpad(context, DPAD_DOWN, 50, 50);      // - Does nothing.          |  - moves selector away from the closed game to avoid opening it.
     pbf_press_dpad(context, DPAD_DOWN, 50, 50);      // - Does nothing.          |  - Press Down a second time in case we drop one.
     pbf_mash_button(context, BUTTON_A, 50);          // - Confirm close game.    |  - opens an app on the home screen (e.g. Online)
-    pbf_press_button(context, BUTTON_HOME, 50, 50);  // - Does nothing.          |  - goes back to home screen.
+    pbf_press_button(context, BUTTON_HOME, 50, 125); // - Does nothing.          |  - goes back to home screen.
+    context.wait_for_all_requests();
```
Mysticial, line 36: This wait doesn't do anything. The commands get queued up on the microcontroller and run with the specified timings.

So in the old code, that delay is 50+50 ticks before the 2nd home press which is almost 1 second.

jw098:
From my testing on SV, this close_game sequence works with and without video input.

PR #422, opened 2024-04-30
Fixed recovery on errors with fly_to_closest_pokecenter_on_map

jw098:
Currently, `fly_to_visible_closest_pokecenter_cur_zoom_level` will re-try if it fails to fly to the pokecenter at a given zoom level. However, on its re-tries, it tries again at the default zoom level, because all it does is close the map and re-open it at the default zoom level. If the initial zoom level was different, this recovery sequence won't work.

My new change pulls this recovery logic into `fly_to_closest_pokecenter_on_map` instead and will ensure the correct zoom level before reattempting.


Gin890:
LGTM! Thanks for the fix!

PR #423, opened 2024-04-30
Refactor material farmer for preparation for integration with item printer

jw098:
This is based off my previous material farmer branch. It has been refactored so that it can be called by the item printer.

I already have a branch where the Item printer automatically calls the material farmer. I'll submit that as a separate pull request later.


Mysticial:
Rebase this since the material farmer got merged.

Mysticial:
Can you split the changes to the ItemPrinterRNG into a separate PR? IOW, keep this PR as refactor only.

jw098:
I have split out the Item Printer changes into its own PR (#432).
This PR is now just the refactor.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
```diff
@@ -0,0 +1,64 @@
...
+#include "NintendoSwitch/Options/NintendoSwitch_GoHomeWhenDoneOption.h"
+#include "PokemonSV/Options/PokemonSV_EncounterBotCommon.h"
+#include "PokemonSV/Options/PokemonSV_SandwichMakerOption.h"
+#include "PokemonSV/Programs/ShinyHunting/PokemonSV_LetsGoTools.h"
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonSV{
+
+struct MaterialFarmerOptions {
```
Mysticial, line 26: If these settings will be duplicated across all material farmer programs, I'm thinking it would be better to pull this out into it's own `BatchOption`. `EncounterBotCommonOptions` is an example of this.

jw098:
I'm not quite done with the switch from struct to Group/BatchOption. I have to think about how to deal with the options that are redundant, such as Notifications and Go Home when Done.

Mysticial:
`SandwichMakerOption` has an example of this for the game language. It's somewhat messy though since this is so far the only time we've had to deal with a redundancy.

The idea is that you pass the option in as a pointer. Whether it's null will determine whether the batch/group will use it versus creating one locally. Then only if it's created locally do you `PA_ADD_OPTION` add the option.

The effect is:

- If no option is passed in, the option will appear inside the box that the `GroupOption` creates. (`BatchOption` doesn't create a box, it just inlines it. But it's functionally the same minus the checkbox that disables the entire batch/group.)
- If an option is passed in, no option will appear inside the box that the `GroupOption` creates. Instead it will be whatever it references (somewhere else in the config).

So the sandwich maker program itself has the language option embedded inside since the sandwich maker option is the only thing that uses it. While the platform bot has the language option at the top (since the program uses it multiple places) and the sandwich section omits it and instead references the top one.

jw098:
I tried changing my code to reduce duplicates similar to SandwichMakerOption. But I'm having a problem with my code causing the program to crash.

The memory dump file indicates that the error is from the Notifications in my member initializer list.
```
MaterialFarmerOptions::MaterialFarmerOptions()
	:
(...)
    , NOTIFICATIONS({
        &NOTIFICATION_STATUS_UPDATE,
        &NOTIFICATION_PROGRAM_FINISH,
        &NOTIFICATION_ERROR_RECOVERABLE,
        &NOTIFICATION_ERROR_FATAL,
    })
(...)
```

I suspect that it's because the member fields are EventNotificationOption& instead of EventNotificationOption.

```
    EventNotificationOption& NOTIFICATION_STATUS_UPDATE;
    EventNotificationOption& NOTIFICATION_PROGRAM_FINISH;
    EventNotificationOption& NOTIFICATION_ERROR_RECOVERABLE;
    EventNotificationOption& NOTIFICATION_ERROR_FATAL;
```

Any suggestions?

On a different note, I seem to have issues with trying to assign new values to any of the option members. I also can't call their copy constructor. If these were intentionally disabled, what was the reason for this?

Mysticial:
Hmm, the notifications are a bit weird. The 4 that are there are almost universal to every program. So can just require them and not bother with the messy optional logic.

jw098:
I fixed the issue. I made a mistake and listed the members in the wrong order in the header.

PR #424, opened 2024-05-01
Add material farmer to item printer

jw098:
Adds a material farming sequence to the item printer. Will automatically fly back and forth between the material farming location and the item printer. It can potentially allow you to use the item printer for days at a time, without intervention. From my testing, the main source of failure/errors is the sandwich maker.

Includes code from #423, and #432 so I recommend merging those first. 

EDIT:
Now only #432 needs to be merged first.

EDIT 2:
#433 needs to be merged first.

EDIT 3:
Now ready for merging.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
+++ b/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
@@ -83,7 +83,32 @@ private:
     );
 };
 
-void run_material_farmer(SingleSwitchProgramEnvironment& env, BotBaseContext& context, MaterialFarmerOptions& options);
+class MaterialFarmer_Descriptor : public SingleSwitchProgramDescriptor{
+public:
+    MaterialFarmer_Descriptor();
+
+    struct Stats : public LetsGoEncounterBotStats{
+        Stats()
+            : m_sandwiches(m_stats["Sandwiches"])
+            , m_autoheals(m_stats["Auto Heals"])
+            , m_game_resets(m_stats["Game Resets"])
+            , m_errors(m_stats["Errors"])
+        {
+            m_display_order.insert(m_display_order.begin() + 2, {"Sandwiches", HIDDEN_IF_ZERO});
+            m_display_order.insert(m_display_order.begin() + 3, {"Auto Heals", HIDDEN_IF_ZERO});
+            m_display_order.insert(m_display_order.begin() + 4, {"Game Resets", HIDDEN_IF_ZERO});
+            m_display_order.insert(m_display_order.begin() + 5, {"Errors", HIDDEN_IF_ZERO});
+        }
+        std::atomic<uint64_t>& m_sandwiches;
+        std::atomic<uint64_t>& m_autoheals;
+        std::atomic<uint64_t>& m_game_resets;
+        std::atomic<uint64_t>& m_errors;
+    };
+    virtual std::unique_ptr<StatsTracker> make_stats() const override;
+};
+
+void run_material_farmer(SingleSwitchProgramEnvironment& env, BotBaseContext& context, 
+    MaterialFarmerOptions& options, MaterialFarmer_Descriptor::Stats& stats);
 
 void run_one_sandwich_iteration(SingleSwitchProgramEnvironment& env, BotBaseContext& context, 
     LetsGoEncounterBotTracker& encounter_tracker, MaterialFarmerOptions& options);
--- a/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
+++ b/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
@@ -24,6 +24,28 @@ namespace PokemonAutomation{
 namespace NintendoSwitch{
 namespace PokemonSV{
 
+
+class MaterialFarmerTools {
+public:
+    struct Stats : public LetsGoEncounterBotStats{
+        Stats()
+            : m_sandwiches(m_stats["Sandwiches"])
+            , m_autoheals(m_stats["Auto Heals"])
+            , m_game_resets(m_stats["Game Resets"])
+            , m_errors(m_stats["Errors"])
+        {
+            m_display_order.insert(m_display_order.begin() + 2, {"Sandwiches", HIDDEN_IF_ZERO});
+            m_display_order.insert(m_display_order.begin() + 3, {"Auto Heals", HIDDEN_IF_ZERO});
+            m_display_order.insert(m_display_order.begin() + 4, {"Game Resets", HIDDEN_IF_ZERO});
+            m_display_order.insert(m_display_order.begin() + 5, {"Errors", HIDDEN_IF_ZERO});
+        }
+        std::atomic<uint64_t>& m_sandwiches;
+        std::atomic<uint64_t>& m_autoheals;
+        std::atomic<uint64_t>& m_game_resets;
+        std::atomic<uint64_t>& m_errors;
+    };
+};
+
 class MaterialFarmerOptions : public GroupOption{
 public:
     MaterialFarmerOptions()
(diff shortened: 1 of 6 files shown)


jw098:
I closed #432, and so those changes were moved into this branch instead. 

As mentioned above, #433 should be merged first.

jw098:
This should now be ready for merging.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
```diff
@@ -83,7 +83,32 @@ class MaterialFarmerOptions : public GroupOption{
     );
 };
 
-void run_material_farmer(SingleSwitchProgramEnvironment& env, BotBaseContext& context, MaterialFarmerOptions& options);
+class MaterialFarmer_Descriptor : public SingleSwitchProgramDescriptor{
```
Mysticial, line 86: This breaks the pattern of descriptor + program.

The only part that's being shared is the stats. So pull that out into this file and leave the descriptor in the original file for the MaterialFarmer program.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -558,12 +563,32 @@ ItemPrinterRNG::ItemPrinterRNG()
...
+        true
+    )
+    , NUM_MATERIAL_FARM_ROUNDS(
+        "<b>Number Material farm rounds:</b><br>"
+        "One round is: Doing one item printer session (as per 'Total Rounds' entered above), "
+        "then farming materials at North Provice area 3 (as per 'Number of sandwich rounds' entered below), "
+        "then flying back to the item printer.",
+        LockMode::UNLOCK_WHILE_RUNNING,
+        3
+    )
```
Mysticial, line 580: This whole thing can probably use a better explanation. If I'm understanding it correctly, it's fundamentally two loops.

The inner loop will repeat the table `TOTAL_ROUNDS` number of times and run the material farmer. The outer loop then wraps all this and runs `NUM_MATERIAL_FARM_ROUNDS`.

What I'm thinking here is to actually throw all the material farming stuff under a `GroupOption` and use its checkbox to enable/disable material farming. Depending on whether it's enabled, we change the description of `TOTAL_ROUNDS` between the old text and the new one you're proposing. Likewise, you won't need to manually show/hide the material farmer options since they'll be greyed out automatically by the GroupOption.

Inside the `GroupOption` for material farming, you can add a `StaticTextOption` at the top that explains the new logic with the material farming being an outer-loop over the whole thing.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -774,6 +811,48 @@ void ItemPrinterRNG::program(SingleSwitchProgramEnvironment& env, BotBaseContext
...
+    assert_16_9_720p_min(env.logger(), env.console);
+    ItemPrinterRNG_Descriptor::Stats& stats = env.current_stats<ItemPrinterRNG_Descriptor::Stats>();
+    env.update_stats();
+
+
+
+    if (!AUTO_MATERIAL_FARMING){
+        run_item_printer_rng(env, context, stats);
+    }
+    else {
```
Mysticial, line 826: nit: The coding convention we normally use is `}else{` and no spaces between `if (){`. Though we aren't super-consistent with adhering to it.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
jw098: I'm not sure on the best way to do this. I previously had the descriptor in the original Material Farmer program. But then I got linker issues. Because MaterialFarmer requires MaterialFarmerTools's functions (e.g. run_material_farmer), but then MaterialFarmerTools requires MaterialFarmer's Stats.

I suppose one solution is to just merge MaterialFarmer and MaterialFarmerTools once again? Put everything in a single file?

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
jw098: I'm envisioning something like this: Outer loop(item printer loop -> material farming loop).

But you're right. It's currently confusing. I'll think about how to reorganize the UI.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
Mysticial: What are the linker issues? Unresolved symbols? Or Duplicate symbols?

jw098: I was able to solve the issue by just moving some of the functions (particularly the ones that used the Stats) from MaterialFarmerTools back to MaterialFarmer.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
jw098: I moved the option that enables Material Farming to the top, to hopefully make it more clear that this an outer loop that wraps the Item printer and material farmer. I also added some `StaticTextOption` that changes text depending on whether material farming is enabled or not.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -25,6 +25,8 @@
 #include "PokemonSV/Inference/Dialogs/PokemonSV_DialogDetector.h"
 #include "PokemonSV/Inference/Overworld/PokemonSV_OverworldDetector.h"
 //#include "PokemonSV/Programs/PokemonSV_GameEntry.h"
+#include "PokemonSV/Programs/Farming/PokemonSV_MaterialFarmer.h"
```
Mysticial, line 28: Programs shouldn't depend on other programs. Why are moving all the tools into the MaterialFarmer program?

Put all the common stuff in MaterialFarmerTools. And have ItemPrinterRNG and MaterialFarmer depend on MaterialFarmerTools.

I don't understand what the errors are.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
Mysticial: No don't do this. Keep the common stuff in MaterialFarmerTools. What errors are you getting?

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
Mysticial: `MaterialFarmer_Descriptor` isn't being shared. Only the stats are . Pull the stats out into the tools file and leave `MaterialFarmer_Descriptor` in the MaterialFarmer program.

jw098: Ok. I understand now. I didn't understand what you meant when you said it the first time.
I now see that the Stats aren't tied to the Descriptor.
I'll delete the errant commit, rebase, and fix it.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
```diff
@@ -24,6 +24,28 @@ namespace PokemonAutomation{
 namespace NintendoSwitch{
 namespace PokemonSV{
 
+
+class MaterialFarmerTools {
+public:
+    struct Stats : public LetsGoEncounterBotStats{
```
Mysticial, line 30: You don't need to wrap it in a class. It's no different than any other class so it can sit outside.

PR #428, opened 2024-05-04
Add recovery routine for detecting sandwich hand

jw098:
If the hand_watcher loses track of the sandwich hand, it searches the whole screen for the hand, instead of just its box.
If that fails, the hand may be at the screen edge. So, move the hand to the middle of the screen and re-try searching the whole screen.
If even that fails, it throws an exception.

I also added test code (commented out) so anyone can test the recovery routine.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_SandwichRoutines.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_SandwichRoutines.cpp
@@ -271,14 +271,52 @@ update the location of the sandwich hand
 */
 bool move_then_recover_sandwich_hand_position(
     const ProgramInfo& info, ConsoleHandle& console, BotBaseContext& context, 
-    SandwichHandType hand_type, SandwichHandWatcher& hand_watcher
+    SandwichHandType& hand_type, SandwichHandWatcher& hand_watcher,
+    AsyncCommandSession& move_session
 ){
 
-    pbf_move_left_joystick(context, 255, 255, TICKS_PER_SECOND*5, 100);
-    pbf_move_left_joystick(context, 0, 128, 100, 100);
-    context.wait_for_all_requests();
+    console.log("center the cursor: move towards bottom right, then left slightly.");
+    int num_ticks_to_move_1 = TICKS_PER_SECOND*4;
+    int num_ticks_to_move_2 = 100;
+
+    // center the cursor
+    if(SandwichHandType::FREE == hand_type){
+        // move to bottom right corner,
+        pbf_move_left_joystick(context, 255, 255, num_ticks_to_move_1, 100);
+        // move to left slightly
+        pbf_move_left_joystick(context, 0, 128, num_ticks_to_move_2, 100);
+        context.wait_for_all_requests();
+    }
+    else if(SandwichHandType::GRABBING == hand_type){
+        // center the cursor while holding the A button, so you don't drop the ingredient.
+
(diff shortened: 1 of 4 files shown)


Mysticial:
I put Gin on this again for first pass since this is almost exclusively on code that was originally Gin's.

Mysticial:
Starting to look at this and the while I don't have any issues with the changes themselves, collectively the entire hand movement function is starting to turn into a bit of a mess.

The initial design (from Gin) was rather experimental to begin with and can probably better described as a prototype that made it into production rather than something that we'd consider to be well designed. Thus a year forward now we're seeing bugs with the design which you're patching up, but leading to an even bigger mess of the original code.

This is probably fine if we expect this to be the last time we touch the hand movement code. But we're easily on the tipping point of "tech debt leading to more bugs causing more tech debt to fix". IOW, I'm willing to allow this for a one-off program within a game where there is a clear "end" and that the code won't be further developed. But if this was infra, I'd block it and force everyone to come up with a better design that's more maintainable.

Let's discuss this in Discord at some point.

jw098:
- `detect` now searches the whole screen if it fails to find the sandwich hand within the box.
- For the recovery sequence, we run `recover_sandwich_hand_position`. This function essentially does the same thing as `process_frame` for now, now that `detect` also searches the whole screen. But it leaves room for more detailed searches in the future (e.g. checking for black outline).

SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_SandwichRoutines.cpp
```diff
@@ -262,6 +263,42 @@ std::string box_to_string(const ImageFloatBox& box){
     return os.str();
 }
 
+/* 
+- center the cursor by moving the cursor to the edge of the screen, then away from the edge.
+- then search the whole screen for the sandwich hand, instead of just the box, and
+update the location of the sandwich hand
+- return true if successful. else throw an exception
+*/
+bool move_then_recover_sandwich_hand_position(
```
Mysticial, line 272: Does this work if you're holding down A to hold an ingredient?

jw098: I'll have to double check later, but I think that it drops the ingredient when it tries the recovery sequence.

jw098: If we want the hand to keep holding down A, do I need to use `move_session.dispatch([&](BotBaseContext& context)`?

Mysticial: Not sure. There's two ways to do this with different levels of elegance and error-proneness.

The elegant but error-prone method is to use `ssf_` and issue a very long A press that returns immediately. Then do all the movement using only `ssf_` touching the joystick buttons only. This effectively "holds" the A button while you do other stuff. But you'll eventually need to issue a cancel to stop the A press.

The less elegant, but safer method is to take whether A is being held down as a parameter. Then on every joystick movement, you either press the joysticks alone, or you use `pbf_controller_state()` to simultaneously press both A and joystick - being careful not to insert any waits that would imply releasing all buttons (including the A).

I *think* Gin's current code is the latter. `ssf_` wasn't available when Gin first write all this. And `ssf` is hard to use if you don't know how it works.

jw098:
I pushed a new update. The recovery sequence continues to hold A when grabbing ingredients.

PR #429, opened 2024-05-05
Update sandwich hand detector so it can also check for the black outline

jw098:
Updated the detector to reduce false positives on lettuce. It checks that the matched waterfill object contains the black outline.
- for waterfill objects that matched the hand template, apply a mask filter to retain all black components
- if this is a true positive, waterfill_sub_binary_image should be a black outline of the hand. 
- if false positive, it will be a pure white image
- then apply the waterfill+template match algorithm on the resulting waterfill_sub_binary_image
- this helps to remove false positives (e.g. lettuce)

SandwichHandLocator::detect will still use the old method of detection, for performance reasons (I didn't test it, but I imagine that adding an extra round of waterfill matching will worsen the lag in an already laggy program).
So, I think that detecting the black outline should only be done for recovery routines.

I recommend merging #428 first.


jw098:
Closing for now, since this doesn't seem to be an issue with the new sandwich hand template.

PR #430, opened 2024-05-06
Another recovery sequence for the sandwich hand, when it gets stuck

jw098:
Another recovery sequence for the sandwich hand, in the case where the hand is "detected", but its movement is stuck.
One situation where this can happen is if the detector false positives on something else (e.g. the lettuce).

The recovery sequence moves the hand into the middle of the screen (goes to corner then the middle), then searches the whole screen for the hand. But instead of just using the normal detector, it will also check for the black outline as well to deal with false positives (see #429).

I recommend merging #429 and #428 before merging this PR.


jw098:
Closing for now, since this doesn't seem to be an issue with the new sandwich hand template.

PR #433, opened 2024-05-14
Move between item printer and material farmer

jw098:
- added functions for moving between item printer and north province 3.
- added more options to material farmer (for testing in dev mode): custom time per sandwich, custom number of forward moves per let's go iteration
- added snapshot on failure
- increase max number of failed attempts

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
@@ -140,7 +140,13 @@ MaterialFarmerOptions::MaterialFarmerOptions(
         "<b>DEV MODE: Time per sandwich:</b><br>Number of minutes before resetting sandwich.",
         LockMode::UNLOCK_WHILE_RUNNING,
         30, 1, 30
-    )    
+    )
+    , NUM_FORWARD_MOVES_PER_LETS_GO_ITERATION(
+        "<b>DEV MODE: Number of forward moves per lets go iteration:</b><br>"
+        "During Let's go autobattling sequence, the number of forward movements before resetting to Pokecenter.",
+        LockMode::UNLOCK_WHILE_RUNNING,
+        13
+    )
     , NOTIFICATION_STATUS_UPDATE(notif_status_update_option == nullptr 
         ? *m_notif_status_update_owner 
         : *notif_status_update_option
@@ -168,6 +174,7 @@ MaterialFarmerOptions::MaterialFarmerOptions(
         PA_ADD_OPTION(SAVE_DEBUG_VIDEO);
         PA_ADD_OPTION(SKIP_WARP_TO_POKECENTER);
         PA_ADD_OPTION(SKIP_SANDWICH);
+        PA_ADD_OPTION(NUM_FORWARD_MOVES_PER_LETS_GO_ITERATION);
     }
     PA_ADD_OPTION(SAVE_GAME_BEFORE_SANDWICH);
     PA_ADD_OPTION(NUM_SANDWICH_ROUNDS);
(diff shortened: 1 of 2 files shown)


SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
```diff
@@ -3,7 +3,7 @@
  *  From: https://github.com/PokemonAutomation/Arduino-Source
  *
  */
-
+#include <QDir>
```
Mysticial, line 6: Why is this needed? We try to keep Qt out of the main programs, though some exceptions are allowed for things that we don't have our own implementation for. (stuff that we should go back an cleanup at some point)

jw098: My bad. It was from when I was experimenting with dumping screenshots. It's not needed. I removed it now.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
```diff
@@ -94,13 +96,36 @@ void lets_go_movement0(BotBaseContext& context);
 
 void lets_go_movement1(BotBaseContext& context);
 
-bool is_sandwich_expired(WallClock last_sandwich_time);
+bool is_sandwich_expired(WallClock last_sandwich_time, int time_per_sandwich);
```
Mysticial, line 99: Specify what units this is in. Or better yet, change it to an `std::chrono` type.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
```diff
@@ -568,6 +594,262 @@ void run_from_battles_and_back_to_pokecenter(SingleSwitchProgramEnvironment& env
...
+        open_map_from_overworld(env.program_info(), env.console, context);
+
+        // change from Paldea map to Blueberry map
+        pbf_press_button(context, BUTTON_L, 50, 300);
+
+        // move cursor to bottom right corner
+        pbf_move_left_joystick(context, 255, 255, TICKS_PER_SECOND*5, 50);
+
+        // move cursor to Blueberry academy fast travel point (up-left)
+        pbf_move_left_joystick(context, 0, 0, 76, 50);
```
Mysticial, line 625: Is this zoom invariant?

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
```diff
@@ -568,6 +594,262 @@ void run_from_battles_and_back_to_pokecenter(SingleSwitchProgramEnvironment& env
...
+    pbf_move_left_joystick(context, 0, 128, 10, 50);
+}
+
+void move_from_item_printer_to_material_farming(SingleSwitchProgramEnvironment& env, BotBaseContext& context){
+    env.console.log("Start moving from item printer to material farming.");
+    move_from_item_printer_to_blueberry_entrance(env, context);
+    fly_from_blueberry_to_north_province_3(env, context);
+}
+
+// assumes you start in the position in fron of the item printer, as if you finished using it.
```
Mysticial, line 730: typo, "in front of"

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
jw098: No. It requires the default zoom level. But this is guaranteed because this routine first closes all menus, then opens the map, which gives us the default zoom level.

PR #437, opened 2024-06-01
Added "Desired Item" column to ItemPrinterRNG

jw098:
- added new column to the table "Desired item", which is a drop-down of high value items from the item printer (e.g. Ability Patch, Master ball, etc.). 
- When the user selects an item, it will then automatically change the date in that row to the correct seed, as well as the correct number of jobs.
- if the user then changes the date (or any other items in the row), the "Desired items" drop down is changed back to "---".
- seeds taken from Anubis's Item printer seeds: https://gist.github.com/Lusamine/112d4230919fadd254f0e6dfca850471
- seeds verified by manual testing

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
@@ -26,6 +26,7 @@
 #include "PokemonSV/Inference/Overworld/PokemonSV_OverworldDetector.h"
 //#include "PokemonSV/Programs/PokemonSV_GameEntry.h"
 #include "PokemonSV_ItemPrinterRNG.h"
+#include <iostream>
 
 namespace PokemonAutomation{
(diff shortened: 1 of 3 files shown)


SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -42,6 +43,53 @@ static const EnumDatabase<ItemPrinterJobs>& ItemPrinterJobs_Database(){
...
+        {ItemPrinterItems::MOON_BALL, "moon-ball", "Moon Ball"},
+        {ItemPrinterItems::SAFARI_BALL, "safari-ball", "Safari Ball"},
+        {ItemPrinterItems::SPORT_BALL, "sport-ball", "Sport Ball"},
+
+    });
+    return database;
+}
+
 ItemPrinterRngRow::~ItemPrinterRngRow(){
     chain.remove_listener(*this);
```
Mysticial, line 94: Need to remove all the listeners you added or it may crash the program.

PR #439, opened 2024-06-02
Update EggFetcher_Core.c

Vardan2210:
Extended the delays in the fly to gate function to combat lag in pokemn SV introduced post the new updates. Also made it so when the program starts the program first automatically flies to the gate inorder to combat any positioning error that may have been caused due to stick drift. Also makes it simpler to launch the program as now the user just needs to be in the crater as opposed to earlier where they had to fly to it and stand on roughly the exact same spot.

### DIFF
--- a/NativePrograms/PokemonSV/Programs/EggFetcher_Core.c
+++ b/NativePrograms/PokemonSV/Programs/EggFetcher_Core.c
@@ -10,7 +10,7 @@
 
 void fly_to_gate(void) {
     pbf_press_button(BUTTON_A, 20, 500);
-    pbf_press_button(BUTTON_Y, 20, 800);
+    pbf_press_button(BUTTON_Y, 20, 1000);
     pbf_move_left_joystick(STICK_MAX, STICK_MAX, 6, 500);
     pbf_mash_button(BUTTON_A, 1000);
 };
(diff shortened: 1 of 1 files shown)


NativePrograms/PokemonSV/Programs/EggFetcher_Core.c
```diff
@@ -47,6 +47,8 @@ int main(void) {
     };
     pbf_press_button(BUTTON_A, 5, 500);
 
+    fly_to_gate();
```
NympheaR, line 50: Sorry I just got started with reviewing. Previously fly_to_gate() had a high fail rate if the player is already at or close to the fly spot as the cursor would snap to the player icon instead of the fly spot. Could you stress test this and if there are fail cases add a move forward sequence (using map marker alignment) so the player is decently away from the fly spot? (this would be useful to implement after the hatching loop as well)

Vardan2210: I have been using this same code, non stop for the past month or so and till now there has been no instance where this failed. It works no matter where you are standing in the general area.

Mysticial:
Closing because MC programs have been deprecated and are no longer supported. We will soon be removing it from the repo entirely.

PR #440, opened 2024-06-03
Automatically adjust delay for ItemPrinterRNG

jw098:
- added ability to automatically adjust the delay for ItemPrinterRNG, by comparing the desired item with the actual print results
- I did not port in Kurt's program. So, this only works for the pre-determined "Desired items". 
- I confirmed empirically that the program is able to correctly detect when it hits the seed, target plus 1, and target minus 1 for all the pre-determined "Desired Items". I didn't do this for the plus 2 or minus 2 cases.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
@@ -707,7 +707,7 @@ void ItemPrinterRNG::run_print_at_date(
 }
 
 void ItemPrinterRNG::adjust_delay(
-    std::array<std::string, 10> print_results, 
+    const std::array<std::string, 10>& print_results, 
     uint64_t seed, ConsoleHandle& console
 ) {
     DistanceFromTarget distance_from_target = get_distance_from_target(print_results, seed, console);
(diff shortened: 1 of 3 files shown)


SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -697,6 +705,377 @@ void ItemPrinterRNG::run_print_at_date(
...
+        new_delay = 0;
+    }
+    DELAY_MILLIS.set((uint16_t) new_delay);
+    console.log("Current delay: " + std::to_string(new_delay));
+    // std::cout << "Current delay:" << new_delay << std::endl;
+    
+}
+
+DistanceFromTarget ItemPrinterRNG::get_distance_from_target(
+    std::array<std::string, 10> print_results,
```
Mysticial, line 749: Always take large objects by reference if they don't need to be copied. (And const-reference if they don't need to be modified)

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -697,6 +705,377 @@ void ItemPrinterRNG::run_print_at_date(
...
+        unknown_seed = true;
+        console.log("Unknown seed. Can't calculate distance from target.");
+        break;
+    }
+
+    if (unknown_seed){
+        return DistanceFromTarget::UNKNOWN;
+    }
+
+    if (print_results == target_result){
```
Mysticial, line 1051: I wonder if we need any leeway for misreads. So rather than requiring that everything matches, we perhaps allow maybe 1 mismatch, but only if there's at least N ingredients in the first place.

It may not be an issue for English, but I can see Chinese being an issue.

PR #441, opened 2024-06-07
Allow user to enable/disable sandwich making for MaterialFarmer

jw098:
- minor UI change for MaterialFarmer
- allows the user to enable/disable sandwich making.
- Sandwich making is disabled by default so that if users use this program without reading the documentation and changing the defaults, it won't burn through all their Herba Mystica (as a side note, I also want to change the default Sandwich recipe so that a Herba mystica recipe isn't the default. But that will come later)

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
@@ -42,6 +42,11 @@ namespace PokemonSV{
 
 using namespace Pokemon;
 
+
+MaterialFarmerOptions::~MaterialFarmerOptions(){
+    ENABLE_SANDWICH.remove_listener(*this);
+}
+
 MaterialFarmerOptions::MaterialFarmerOptions(
     OCR::LanguageOCROption* language_option, GoHomeWhenDoneOption* go_home_when_done_option,
     EventNotificationOption* notif_status_update_option, EventNotificationOption* notif_program_finish_option, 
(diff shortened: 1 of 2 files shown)


pifopi (approved):

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
```diff
@@ -24,12 +24,12 @@ namespace PokemonAutomation{
 namespace NintendoSwitch{
 namespace PokemonSV{
 
-class MaterialFarmerOptions : public GroupOption{
+class MaterialFarmerOptions : public GroupOption, public ConfigOption::Listener{
 public:
+    ~MaterialFarmerOptions();
     MaterialFarmerOptions()
         : MaterialFarmerOptions(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr)
```
Mysticial, line 31: You don't need this. `std::unique_ptr` default initializes to null.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
```diff
@@ -121,18 +125,18 @@ MaterialFarmerOptions::MaterialFarmerOptions(
...
-        LockMode::LOCK_WHILE_RUNNING,
+        LockMode::UNLOCK_WHILE_RUNNING,
         false
     )
-    , SKIP_SANDWICH(
-        "<b>DEV MODE: Skip making sandwich:</b><br>"
-        "This is for debugging the program without waiting for sandwich making.",
+    , ENABLE_SANDWICH(
+        "<b>Enable Sandwich making:</b><br>"
+        "This is for boosting spawn rates of specific Pokemon.",
```
Mysticial, line 139: You don't need a separate option to toggle the entire sandwich part. The `SandwichMakerOption` inherits from `GroupOption` which supports adding a checkbox to toggle the entire thing.

Lemme make a small change to `SandwichMakerOption` to pass this feature down.

The same thing applies to the entire material farmer in the Item Printer. I already merged it, but it uses a separate option to toggle it. You can also just defer to the `GroupOption` to show the checkbox instead.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
Mysticial: Just do `MaterialFarmerOptions() = default`. (Which is the same thing as `MaterialFarmerOptions(){}`)

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
Mysticial: Lemme think about this. Since I haven't used the program yet. The reason why my line of thought went in this this direction is because I feel the sandwich/material farmer are add-ons to existing programs - not something that fundamentally changes their behavior.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.h
jw098: Sorry for all the questions, but I still don't understand what you're suggesting. I tried changing the constructor on line 30 to `MaterialFarmerOptions() = default`, but then I get a compile error.  "function was implicitly deleted because 'PokemonAutomation::NintendoSwitch::PokemonSV::MaterialFarmerOptions' has an uninitialized data member"

Currently, `MaterialFarmerOptions()` delegates to the constructor with all the pointers, which is the reason for `MaterialFarmerOptions(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr)`. How would setting it to `default` work in this case?

Mysticial: Oh you're right. There's a bunch of other crap that needs to be initialized. Somehow I thought that other constructor was empty.

Nevermind then. It's fine as is.

SerialPrograms/Source/PokemonSV/Programs/Farming/PokemonSV_MaterialFarmerTools.cpp
Mysticial: I'll merge this just to get it out of the way. But will probably tweak things later.

PR #443, opened 2024-06-08
Added Normal Encounter (non-shiny) sandwich to Sandwich Maker

jw098:
- this is mainly helpful for the Material farmer for farming Happiny dust.
- changed the default sandwich to the normal (non-shiny) sandwich, so users don't accidentally use up Herba Mystica when they weren't intending to shiny hunt.
- if there is a need for to add non-shiny encounter sandwiches for the rest of the types, I can add them later. But I don't see a use for them at this time. So, right now, the BaseRecipe::non_shiny is basically hard-coded to SandwichRecipe::non_shiny_normal.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Options/PokemonSV_SandwichMakerOption.cpp
+++ b/SerialPrograms/Source/PokemonSV/Options/PokemonSV_SandwichMakerOption.cpp
@@ -18,6 +18,7 @@ namespace PokemonSV{
 //  Map for standard+type recipes taking Base + Type to find SandwichRecipe
 const std::map<std::pair<BaseRecipe, PokemonType>, SandwichRecipe>& PREMADE_SANDWICH_TYPE(){
     static const std::map<std::pair<BaseRecipe, PokemonType>, SandwichRecipe> map{
+        {{BaseRecipe::non_shiny, PokemonType::normal},  SandwichRecipe::non_shiny_normal},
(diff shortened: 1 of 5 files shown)


SerialPrograms/Source/PokemonSV/Options/PokemonSV_SandwichMakerOption.cpp
```diff
@@ -313,14 +315,15 @@ SandwichMakerOption::SandwichMakerOption(
...
+            {BaseRecipe::non_shiny,  "non-shiny",    "Normal Encounter (non-shiny)"},
             {BaseRecipe::shiny,     "shiny",    "Sparkling + Title + Encounter"},
             {BaseRecipe::huge,      "huge",     "Sparkling + Title + Humungo"},
             {BaseRecipe::tiny,      "tiny",     "Sparkling + Title + Teensy"},
             {BaseRecipe::paradox,   "paradox",  "Title + Encounter + Humungo/Teensy: Paradox-specific"},
             {BaseRecipe::custom,    "custom",   "Custom Sandwich"},
         },
         LockMode::LOCK_WHILE_RUNNING,
-        BaseRecipe::shiny
+        BaseRecipe::non_shiny
```
Mysticial, line 326: This will also change the defaults for the platform and scatterbug programs. Maybe we should take the default as a parameter.

PR #446, opened 2024-06-19
Warn user about overlapping Item/Ball bonuses

jw098:


SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -207,6 +211,44 @@ void ItemPrinterRNG::value_changed(void* object){
         AFTER_ITEM_PRINTER_DONE_EXPLANATION.set_text("Then stop the program.");
         MATERIAL_FARMER_DISABLED_EXPLANATION.set_text("<br>To enable the Material Farmer, enable \"Automatic Material Farming\" above");
     }
+
+    if (overlapping_bonus()){
+        OVERLAPPING_BONUS_WARNING.set_text(
```
Mysticial, line 216: Instead of setting and clearing the text, what about showing and hiding the entire option. This will eliminate the extra newline that would be there with an empty message.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -207,6 +213,41 @@ void ItemPrinterRNG::value_changed(void* object){
...
+
+// - return true if the user is trying to trigger another Ball/Item bonus without
+// using up the previous bonus. Doing this can mess up rewards for subsequent prints, 
+// since a ball bonus interferes with activating an item bonus and vice versa.
+// - each Bonus lasts for 10 prints.
+bool ItemPrinterRNG::overlapping_bonus(){
+    // for each row in table. if ball/item bonus, ensure that sum of prints in subsequent rows >=10 before the next bonus, or end of the table.
+    uint16_t total_prints_since_last_bonus = 10;
+    for (std::shared_ptr<EditableTableRow> table_row : TABLE0.current_refs()){
+        std::shared_ptr<ItemPrinterRngRow> row = std::static_pointer_cast<ItemPrinterRngRow>(table_row);
```
Mysticial, line 235: It's actually safe to take the reference directly here since the lifetime is already preserved in the outer loop.

`ItemPrinterRngRow& row = static_cast<ItemPrinterRngRow&>(*table_row);`

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -207,6 +213,41 @@ void ItemPrinterRNG::value_changed(void* object){
...
+        if (row->desired_item == ItemPrinter::PrebuiltOptions::BALL_BONUS || row->desired_item == ItemPrinter::PrebuiltOptions::ITEM_BONUS){
```
Mysticial, line 236: We don't need to do this now, but I could extend the seed calculator to return whether it generated a bonus and go from that. Though it can get a bit flaky with missed frames and such. (which doesn't happen with the built-ins because they have consecutive frames of bonuses)

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
jw098: I can't get your suggestion to compile.
From my reading, shared_ptr can only be cast with static_pointer_cast.

PR #452, opened 2024-06-24
Use English as OCR language for item printer jobs detector

jw098:


SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterJobsDetector.h
```diff
@@ -42,7 +42,7 @@ class ItemPrinterJobsDetector{
 
 private:
     Color m_color;
-    Language m_language;
+    Language m_language;  // language isn't really used, since it default to English
```
Mysticial, line 45: Then let's delete this field then.

PR #453, opened 2024-06-25
Detect quantity of Happiny Dust, to trigger Material farming from Item printer

jw098:
- add the option for users to specify a minimum Happiny Dust threshold, below which Material farming will be triggered from the Item printer
- This also helps ensure that other materials don't drop below this number as well, since the Item printer prioritizes using up materials with the highest quantity.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
+++ b/SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
@@ -29,7 +29,9 @@ MaterialNameReader& MaterialNameReader::instance(){
     return reader;
 }
 
-
+// MaterialNameReader has a very limited dictionary,
+// so it can only reliably read the material names with 68% value
+// (i.e. Ditto Goo, Happiny Dust, Magby Hair, Beldum Claw)
 MaterialNameReader::MaterialNameReader()
     : SmallDictionaryMatcher("PokemonSV/ItemPrinterMaterialOCR.json")
 {}
@@ -230,7 +232,7 @@ int8_t ItemPrinterMaterialDetector::find_happiny_dust_row_num(
             && detect_material_name(console, context, value_68_row_num) == "happiny-dust"
         ){  
             // found screen and row number with Happiny dust.
-            std::cout << "Happiny dust found. Row number: " << std::to_string(value_68_row_num) << std::endl;
+            // std::cout << "Happiny dust found. Row number: " << std::to_string(value_68_row_num) << std::endl;
             return value_68_row_num;
         }
         // keep searching for Happiny dust
@@ -245,6 +247,10 @@ int8_t ItemPrinterMaterialDetector::find_happiny_dust_row_num(
 
 }
 
+// detects the material name at the given row_num
+// MaterialNameReader has a very limited dictionary,
+// so it can only reliably read the material names with 68% value
+// (i.e. Ditto Goo, Happiny Dust, Magby Hair, Beldum Claw)
 std::string ItemPrinterMaterialDetector::detect_material_name(
     ConsoleHandle& console, 
     BotBaseContext& context,
--- a/SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
+++ b/SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
@@ -189,14 +189,16 @@ int8_t ItemPrinterMaterialDetector::find_happiny_dust_row_index(
     int8_t value_68_row_index;
     for (size_t c = 0; c < 30; c++){
         context.wait_for_all_requests();
-        value_68_row_index = find_material_value_row_index(dispatcher, console, context, 68);
-        if (value_68_row_index != -1 
-            && detect_material_name(console, context, value_68_row_index) == "happiny-dust"
-        ){  
-            // found screen and row number with Happiny dust.
-            // std::cout << "Happiny dust found. Row number: " << std::to_string(value_68_row_index) << std::endl;
-            return value_68_row_index;
+        std::vector<int8_t> value_68_row_index_list = find_material_value_row_index(dispatcher, console, context, 68);
+        for (size_t i = 0; i < value_68_row_index_list.size(); i++){
+            value_68_row_index = value_68_row_index_list[i];
+            if (detect_material_name(console, context, value_68_row_index) == "happiny-dust"){  
+                // found screen and row number with Happiny dust.
+                // std::cout << "Happiny dust found. Row number: " << std::to_string(value_68_row_index) << std::endl;
+                return value_68_row_index;
+            }
         }
+        
         // keep searching for Happiny dust
         pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
     }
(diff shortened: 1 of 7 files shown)


SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -629,6 +704,92 @@ void ItemPrinterRNG::run_item_printer_rng(
...
+        case 3:{
+            env.log("Detected material selection.");
+            ItemPrinterMaterialDetector detector(COLOR_RED);
+            // Press the right button 6 times to see the Happiny Dust in the material selector.
+            pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
+            pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
+            pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
+            pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
+            pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
+            pbf_press_dpad(context, DPAD_RIGHT, 20, 30);
```
Mysticial, line 773: What happens if Happiny dust is not visible after 6 scrolls? Say you're missing most of the earlier ones so that 6 scrolls past it.

jw098: True. I forgot about this. 
Number recognition is faster and more reliable right?
I'm thinking we use number recognition to find materials with 68% value, of which there are only 4 materials according to Serebii (Ditto Goo, Happiny Dust, Magby hair, Beldum Claw). Then use string recognition.

Mysticial: Number is worse because there's no error correction. Strings have error correction since we can compare against all the possible string values. And there's more characters to work with.

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.h
```diff
@@ -0,0 +1,133 @@
...
+    ImageFloatBox get_material_quantity_box(int8_t row_num) const;
+
+    ImageFloatBox get_material_value_box(int8_t row_num) const;
+
+    ImageFloatBox get_material_name_box(int8_t row_num) const;
+
+private:
+    Color m_color;
+    Language m_language;
+    ImageFloatBox m_box_1_mat_value;
```
Mysticial, line 96: You don't need to do this. Just do an array. The items are also evenly spaced out vertically. So you don't need to hard-code the coordinates. Given the first one and the space between them, you can use a simple loop to compute the rest.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -156,12 +158,32 @@ ItemPrinterRNG::ItemPrinterRNG()
...
     )
+    , MIN_HAPPINY_DUST(
+        "<b>Minimum Happiny Dust:</b><br>"
+        "Run the material farmer before the number of Happiny Dust drops "
+        "below this number.<br>"
+        "This ensures no other material drops below this number. "
+        "If a material starts below this threshold, it remains there.<br>"
+        "Changes to this number only take place after returning to "
+        "the item printer, after material farming.",
+        LockMode::UNLOCK_WHILE_RUNNING, 100,
```
Mysticial, line 184: I dunno if setting it as low as 100 is necessarily the # since the auto-select always uses the ones with the most quantity first. So if you have larger quantities of non-Happiny dust that are harder to collect, it will consume those as well.

But I'm not sure how much material is used for making TMs. Maybe 100 is enough for most purposes.

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+// return row number where Happiny dust is located on screen
+// keep pressing DPAD_RIGHT until Happiny dust is on screen
+// check each row on the screen for Happiny Dust
+int8_t ItemPrinterMaterialDetector::find_happiny_dust_row_index(
+    AsyncDispatcher& dispatcher,
+    ConsoleHandle& console, BotBaseContext& context
+) const{
+    int8_t value_68_row_index;
+    for (size_t c = 0; c < 10; c++){
```
Mysticial, line 190: 10 is the max # of times to fully page down?

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
```diff
@@ -0,0 +1,289 @@
...
+// return row number that matches given material_value. else return -1
+int8_t ItemPrinterMaterialDetector::find_material_value_row_index(
+    AsyncDispatcher& dispatcher,
+    ConsoleHandle& console, 
+    BotBaseContext& context,
+    int16_t material_value
+) const{
+    context.wait_for_all_requests();
+    VideoSnapshot snapshot = console.video().snapshot();
+    for (int8_t i = 0; i < 10; i++){
```
Mysticial, line 260: I wonder if it's worth parallelizing this since you're running 10 of them sequentially. Not sure how long this takes on a slower computer.

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
```diff
@@ -0,0 +1,289 @@
...
+        int16_t value = read_number(console, dispatcher, snapshot, m_box_mat_value[i]);
+        if (value == material_value){
+            return i;
```
Mysticial, line 263: What happens if there are multiple 68% materials on the same screen?

jw098: My bad. I keep forgetting that there can be missing materials from the list. I'll make the function return a vector instead.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
jw098: TMs usually use 3 for a given material. Occasionally up to 8. 
https://bulbapedia.bulbagarden.net/wiki/TM_Material

I can change the default to 400.

If you run the material farmer for 30 minutes, you'll get around 450 Happiny Dust. So a threshold of around 400 is reasonable. If you run the material for 60 minutes, you'll get around 900 Happiny Dust, so a threshold of around 100 would be better in that case.

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
jw098: No. There are around 181 materials. So should take 18-19 times to fully page down.

But Happiny Dust is number 73, so should take 8 times to page down. As far as I know, the items are always in the same order. 

I'm fine with increasing the max number of times to scroll, just to account for button drops, or in case the material is in a different order than expected.

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
jw098: I'm also not sure if it's worth it. But I guess it couldn't hurt.

PR #459, opened 2024-07-10
Improve performance of Material Detector

jw098:
- use Waterfill OCR instead of straight OCR. I tested it for number 1->999 for both white and black text.
- since we only need 1 thread for Waterfill OCR, I also parallelized detection of material value. I tested it on my old laptop, it runs well, and runs faster than before.


SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterMaterialDetector.cpp
```diff
@@ -84,101 +85,31 @@ int16_t ItemPrinterMaterialDetector::read_number(
...
-
-    std::vector<std::pair<ImageRGB32, size_t>> filtered = to_blackwhite_rgb32_range(
+    ImageRGB32 filtered = to_blackwhite_rgb32_range(
         cropped,
-        filters
+        0xff000000, 0xff808080,
+        true
     );
+    // filtered.save("DebugDumps/test-one-filter-1.png");
+    bool is_dark_text_light_background = image_stats(filtered).average.sum() > 500;
```
jw098, line 96: I'm not sure if this is the best way to check if the text is dark vs light. Happy to hear any suggestions.

Mysticial: 500 feels too high. IIRC, I've had issues detecting whites in the past with 500 on my "bad" capture card.

What are the actual values for dark vs. light? Maybe pick the mid-point?

jw098: Dark text light background is usually >600. Light text dark background is usually <200.
I've changed it to 400.  
Also, I do apply a binary filter first, so that should hopefully help as well.

PR #460, opened 2024-07-12
Confirm sandwich ingredient added, to avoid button drops.

jw098:
Use icon detection to confirm that the selected ingredients are actually added to ensure we didn't drop any buttons.

Since we can confirm that ingredients are added, we can also loop A button presses when multiple of a given ingredient is required. Because if you run out of an ingredient, it will detect that nothing has been added and throw an exception.

Also, since the ingredient reader now needs to also read the confirmation icons, I refactored the reader to contain all the boxes for all rows/columns, not just for one row.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_IngredientSession.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_IngredientSession.cpp
@@ -14,6 +14,9 @@
 #include "PokemonSV/Resources/PokemonSV_Ingredients.h"
 #include "PokemonSV_IngredientSession.h"
 #include "Common/Cpp/PrettyPrint.h"
+#include <iostream>
+using std::cout;
+using std::endl;
 
 namespace PokemonAutomation{
 namespace NintendoSwitch{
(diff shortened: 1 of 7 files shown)


SerialPrograms/Source/PokemonSV/Programs/Sandwiches/PokemonSV_IngredientSession.cpp
```diff
@@ -31,16 +34,16 @@ IngredientSession::IngredientSession(
...
+    , m_type(type)
+    , m_num_confirmed(0)
     , m_arrow(COLOR_CYAN, GradientArrowType::RIGHT, {0.02, 0.15, 0.05, 0.80})
 {
-    for (size_t c = 0; c < INGREDIENT_PAGE_LINES; c++){
-        m_ingredients.emplace_back(type, c, COLOR_CYAN);
-        m_ingredients.back().make_overlays(m_overlays);
-    }
+    SandwichIngredientReader reader(m_type, COLOR_CYAN);
+    reader.make_overlays(m_overlays);
```
Mysticial, line 42: I don't think this works the way you think it does. Once `reader` is destructed, it removes all the boxes from the overlay.

jw098: I just tested it. And the boxes still show up on screen. (and the boxes go away if I remove that code)
Why does my code work?

Mysticial: Looks like I confused `VideoOverlaySet` with the `VideoOverlay` itself. The latter holds references to the boxes while the former owns the boxes. So when `VideoOverlaySet` is destructed, it automatically removes the boxes it holds from the `VideoOverlay`.

You're good.

PR #462, opened 2024-07-23
Item Printer Auto mode

jw098:
- added a new mode for the Item printer, where instead of users manually entering in all the print jobs, users can just specify which item they want and in what quantity. I call it Auto mode (vs Standard mode), but I'm happy to hear other name suggestions
- it will automatically run the material farmer when Happiny Dust is less than 400.
- it uses a map of all cumulative prize and quantity, so we know how many prints to do


SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
```diff
@@ -604,6 +618,194 @@ void ItemPrinterRNG::print_again(
...
+    // Check material quantity when:
+    // - once when first starting the item printer
+    // - before starting material farming. If still have material, 
+    // can keep using item printer. But this check is only done once, 
+    // until you farm materials again.
+    // - when back from material farming
+    bool done_one_last_material_check_before_mat_farming = false;
+    uint32_t material_farmer_jobs_period = calc_num_jobs_using_happiny_dust(env, context, min_happiny_dust);   
+    bool have_cleared_out_bonus = false;
+    std::map<std::string, uint16_t> obtained_prizes;
```
jw098, line 655: Do you think a map or unordered map would be better here? Or maybe it doesn't matter? The number of items in the map can range anywhere from 5-10 to 120, depending on how many different desired items are selected.

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterPrizeReader.cpp
```diff
@@ -117,7 +124,76 @@ std::array<std::string, 10> ItemPrinterPrizeReader::read(
...
+    for (int8_t i = 0; i < total_rows; i++){
+        // ImageRGB32 filtered = to_blackwhite_rgb32_range(
+        //     extract_box_reference(screen, m_boxes_bonus_quantity[i]),
+        //     0xff808000, 0xffffffff,
+        //     true
+        // );
+        // filtered.save("DebugDumps/test"+ std::to_string(i) +".png");   
+
+        tasks[i] = dispatcher.dispatch([&, i]{
+            if (total_average_sum_normal > total_average_sum_bonus){
```
jw098, line 156: Basic programming question: Is there a more elegant way of formatting the if statement here?

I was thinking of something like this. But without `boxes` copying the whole array.
```
std::array<ImageFloatBox, 10> boxes = m_boxes_bonus_quantity;
if (total_average_sum_normal > total_average_sum_bonus){
	boxes = m_boxes_normal_quantity;
}

for (int8_t i = 0; i < total_rows; i++){
       tasks[i] = dispatcher.dispatch([&, i]{
            results[i] = read_number(logger, screen, boxes[i]);            
        });
    }
```

pifopi: you can make boxes be a reference ?
std::array<ImageFloatBox, 10>**&** boxes = m_boxes_bonus_quantity;

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
pifopi: If ordering doesn't matter, unordered map are better. But for such small things, it won't matter much in most cases

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterPrizeReader.cpp
jw098: When I try making `boxes` a reference, I get this error: 

C2440: 'initializing': cannot convert from 'const std::array<PokemonAutomation::ImageFloatBox,10>' to 'std::array<PokemonAutomation::ImageFloatBox,10> &'

`const std::array<ImageFloatBox, 10>& boxes = m_boxes_bonus_quantity;` doesn't work either: 

error C2678: binary '=': no operator found which takes a left-hand operand of type 'const std::array<PokemonAutomation::ImageFloatBox,10>' (or there is no acceptable conversion)

Also, would a reference even work? Since references can't be reassigned? Once I assign `boxes = m_boxes_bonus_quantity;`, then I can't reassign it to `m_boxes_normal_quantity`, if the if-statement is true, correct?

pifopi: oh yes you're right, must have been tired this morning. You could do a method that return "const std::array<ImageFloatBox, 10>&"

Then inside it, based on "total_average_sum_normal" and "total_average_sum_bonus", you return one of the two array ? I guess it could be a lambda too. But honeslty, i feel current version is okay ?

jw098: Thanks! Yeah, a separate function/lambda works. It looks like a ternary operator works as well.
If you feel it's fine as it is, then I'll just leave it.

Mysticial: I would use the ?: ternary operator.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNGTable.cpp
```diff
@@ -134,6 +134,77 @@ std::vector<std::unique_ptr<EditableTableRow>> ItemPrinterRngTable::make_default
...
+std::vector<std::string> ItemPrinterDesiredItemTable::make_header() const{
+    return std::vector<std::string>{
+        "Desired Item",
+        "Quantity",
+    };
+}
+std::vector<std::unique_ptr<EditableTableRow>> ItemPrinterDesiredItemTable::make_defaults(){
+    std::vector<std::unique_ptr<EditableTableRow>> ret;
+    ret.emplace_back(std::make_unique<ItemPrinterDesiredItemRow>(*this, ItemPrinter::PrebuiltOptions::ABILITY_PATCH, 999));
+        ret.emplace_back(std::make_unique<ItemPrinterDesiredItemRow>(*this, ItemPrinter::PrebuiltOptions::EXP_CANDY, 999));
```
Mysticial, line 203: Indentation

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.h
```diff
@@ -95,14 +130,6 @@ class ItemPrinterRNG : public SingleSwitchProgramInstance, public ConfigOption::
     GoHomeWhenDoneOption GO_HOME_WHEN_DONE;
     BooleanCheckBoxOption FIX_TIME_WHEN_DONE;
 
-//    BooleanCheckBoxOption AUTO_MATERIAL_FARMING;
-//    SimpleIntegerOption<uint16_t> NUM_ROUNDS_OF_ITEM_PRINTER_TO_MATERIAL_FARM;
-    
-//    StaticTextOption MATERIAL_FARMER_DISABLED_EXPLANATION;
-    enum class MaterialFarmerTrigger{
```
Mysticial, line 102: Is the new mode a new mode? Or is it replacing the existing auto-farming mode?

SerialPrograms/Source/PokemonSV/Inference/ItemPrinter/PokemonSV_ItemPrinterPrizeReader.cpp
```diff
@@ -117,7 +124,75 @@ std::array<std::string, 10> ItemPrinterPrizeReader::read(
 }
 
 
+std::array<int16_t, 10> ItemPrinterPrizeReader::read_quantity(
```
Mysticial, line 127: Given how big and invasive this PR is, let's break it up. Can start by splitting out this quantity reader.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.h
jw098: Auto mode is a new mode. The existing auto-farming mode is still there. I just moved the MaterialFarmerTrigger enum class higher up in the file, for reasons I can't remember. We can move it back if you want.

SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNG.cpp
Mysticial: If performance isn't important, I usually prefer map over unordered map because of the consistent sorting. Though the converse can be true if you don't want people to accidentally become dependent on the order.

I think some of the higher level languages (was it Javascript?) or something wanted to randomize the iteration order of hash tables to prevent this specifically. lol

Mysticial:
Could probably give better names to the two tables. But I won't block this any further.

PR #463, opened 2024-07-27
Direction detector

jw098:
This can detect and change player direction, based on detecting the N symbol in the minimap.

At first, I tried waterfilling the entire green portion of the compass, then the N. But then I found that directly waterfilling the N works just as well (this is why I made `match_filters_by_waterfill`, but it ended up being unused in the end). It will only detect the N symbol if it's a certain distance away from the radar ball, which should help to minimize false positives.


SerialPrograms/Source/CommonFramework/ImageTools/WaterfillUtilities.h
```diff
@@ -61,6 +62,15 @@ bool match_template_by_waterfill(
     double rmsd_threshold,
     std::function<bool(Kernels::Waterfill::WaterfillObject& object)> check_matched_object);
 
+// Similar to match_template_by_waterfill(), except instead of matching by template,
+// matches by filters, and returns a vector of ImageFloatBox
+std::vector<ImageFloatBox> match_filters_by_waterfill(
```
Mysticial, line 67: So this function is identical to `match_template_by_waterfill()`, but without the template check?

jw098: That's right.
I didn't end up using it in the end, so we can delete if you think it's not useful/redundant.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.h
```diff
@@ -0,0 +1,68 @@
+/*  Fast Travel Detector
```
Mysticial, line 1: Copy-paste. lol

SerialPrograms/Source/CommonFramework/ImageTools/WaterfillUtilities.h
Mysticial: Yeah let's get rid of it for now since it's basically copy-paste of the original. If we need it, we can try to merge them.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
```diff
@@ -0,0 +1,169 @@
...
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonSV{
+
+
+class DirectionMatcher : public ImageMatch::WaterfillTemplateMatcher{
+public:
+    
+    DirectionMatcher() : WaterfillTemplateMatcher(
+        "PokemonSV/Map/DirectionIcon-Template.png", Color(0, 100, 30), Color(50, 230, 85), 50
```
Mysticial, line 37: Don't forget to PR this file.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    const double min_object_size = 50.0;
+    const double rmsd_threshold = 60.0;
+
+    const double screen_rel_size = (screen.height() / 1080.0);
+    const size_t min_size = size_t(screen_rel_size * screen_rel_size * min_object_size);
+
+    std::pair<double, double> north_location(0, 0);
+    std::pair<double, double> zero_coord(0.9065, 0.8335);
```
Mysticial, line 74: Is it a good idea to hard-code the coordinate of the radar like this? Since some capture cards may shift by a few pixels.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
```diff
@@ -0,0 +1,169 @@
...
+            double x1 = (object.center_of_gravity_x() + pixel_search_area.min_x) / (double)screen.width();
+            double y1 = (object.center_of_gravity_y() + pixel_search_area.min_y) / (double)screen.height();
+            // relative coord wrt minimap radar
+            double x2 = x1 - zero_coord.first;
+            double y2 = zero_coord.second - y1; // subtract y from zero_coord since I want north to be positive
+            // absolute coord wrt minimap radar, scaled to screen size
+            double x3 = x2 * (double)screen.width() / screen_rel_size;
+            double y3 = y2 * (double)screen.height() / screen_rel_size;
+
+            double hypotenuse = std::hypot(x3, y3);
```
Mysticial, line 95: I wonder if we should assert a video aspect ratio here.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include "CommonFramework/Logging/Logger.h"
+#include "CommonFramework/Tools/DebugDumper.h"
+#include "CommonFramework/Tools/VideoResolutionCheck.h"
+#include "NintendoSwitch/Commands/NintendoSwitch_Commands_PushButtons.h"
+#include "Kernels/Waterfill/Kernels_Waterfill_Types.h"
+#include "PokemonSV/Inference/Overworld/PokemonSV_OverworldDetector.h"
+#include "PokemonSV_DirectionDetector.h"
+#include <cmath>
+#include <iostream>
+#ifndef M_PI
```
pifopi, line 25: why is this needed ? It should be included with cmath ? (if it's not, try defining _USE_MATH_DEFINES)

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
pifopi: https://github.com/PokemonAutomation/Packages/pull/29

pifopi (approved):

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
```diff
@@ -71,7 +75,16 @@ std::pair<double, double> DirectionDetector::locate_north(const ImageViewRGB32&
     const size_t min_size = size_t(screen_rel_size * screen_rel_size * min_object_size);
 
     std::pair<double, double> north_location(0, 0);
-    std::pair<double, double> zero_coord(0.9065, 0.8335);
+    OverworldDetector detector;
+    std::pair<double, double> zero_coord = detector.locate_ball(screen); // 0.9061, 0.8328
+    if (zero_coord.first > 0.91 || zero_coord.first < 0.90 || 
+        zero_coord.second > 0.84 || zero_coord.second < 0.82)
+    {
+        console.log("Unable to locate the overworld radar ball, falling back on hard coded location.");
```
Mysticial, line 81: Make this log line either red or orange in color so it's easier to see when scrolling through.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_OverworldDetector.cpp
```diff
@@ -124,6 +142,71 @@ bool OverworldDetector::detect_ball(const ImageViewRGB32& screen) const{
     return false;
 }
 
+std::pair<double, double> OverworldDetector::locate_ball(const ImageViewRGB32& screen) const{
```
Mysticial, line 145: Do we still need the `detect_ball()` function above? Can probably delete that and have the caller call `locate_ball()` instead.

jw098: `detect_ball` has looser requirements than `locate_ball`. Since `locate_ball` requires a min_object_size of 150, while `detect_ball` has no requirement. For example, this means `detect_ball` will return true even if it only detects the yellow arrow, and not the ball itself.

This is more of a theoretical concern, though. I believe `locate_ball` should be able to reliably detect the overworld. Thoughts?

jw098: Another factor is that I'm actually thinking of reducing the RMSD threshold for `locate_ball`, down to 40. Since the yellow arrow has an RMSD of 60, while the ball itself usually has an RMSD < 20.

Mysticial: IIRC, the only other user of the radar ball is the overworld detector. So that's the case we need to make sure still works. The corner case being if you're standing next to an event den, the sparkling animation messes things up.

I don't remember the exact details of the overworld detector, but it uses some combination of the map not moving and the radar ball to detect if you're in the overworld. One of them is a fallback for the other in case the sparkling den information messes things up.

So as long as the overworld detector still works, we should be good. Though radar ball has two shades of yellow which is somewhat sensitive to capture card distortion. That may be wide the filter list is so long.

jw098: Yeah. I can do some basic testing to make sure the overworld detector still works (including near sparkling dens). I still worry about weird corner cases (e.g. different capture cards) where the old detection method works (with detect_ball), but the new one doesn't (locate_ball).

Are you ok with this risk? Or should we just leave it as is, with detect using detect_ball instead of locate_ball.

Mysticial: Yeah that's fine. No real reason to fork that much code if that's being used for the same exact thing. If it breaks the overworld detector, we should know pretty easily.

jw098: I decided to add a flag `strict_requirements`. When set to true, it uses a tighter RMSD threshold and has min size requirements for waterfill objects. When set to false, it's more similar to the old `detect_ball` function, where the RMSD threshold is looser and there are no min size requirements.

SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
jw098: I tried using _USE_MATH_DEFINES instead of manually defining M_PI. It seems to be causing the Windows 5.12.12 build to fail.

```D:\a\Arduino-Source\Arduino-Source\Arduino-Source\SerialPrograms\Source\PokemonSV\Inference\Overworld\PokemonSV_DirectionDetector.cpp(129,52): error C2065: 'M_PI': undeclared identifier [D:\a\Arduino-Source\Arduino-Source\Arduino-Source\SerialPrograms\bin\SerialPrograms.vcxproj]```

Any ideas on how to fix this?

PR #466, opened 2024-08-25
Autostory

jw098:
This is quite a large PR. Let me know if you want it broken up.
I updated/added a few detectors/readers
- WipeoutDetector
- WhiteButtonDetector for both L Stick and A button with dark background
- MenuOptionReader along with a menu option session for changing the settings
- PokemonMovesReader
- TutorialDetector

Credits to Nymphea for the majority of the autostory code.


jw098:
I added a few commits to refactor `test_segments`, to use `std::vector<std::function<void()>>` instead of a large switch statement.

Mysticial:
Without having looked into this in detail yet, let's break out all the detectors into separate PRs if they're able to stand alone.

jw098:
I'll submit AutoStoryTools as its own PR once #482 and #483 are merged, since it depends on those.

PR #467, opened 2024-08-25
Python script to generate OCR json files

jw098:


pifopi:
I think @denvoros have something for this already ?

PR #471, opened 2024-09-01
Add MenuOption to change the menu options/settings.

jw098:


SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_MenuOption.h
```diff
@@ -0,0 +1,83 @@
...
+public:
+    ~MenuOption();
+    MenuOption(
+        ConsoleHandle& console, BotBaseContext& context,
+        Language language
+    );
+
+    // change all settings as per options
+    void set_options(
+        std::vector<std::pair<MenuOptionItemEnum, std::vector<MenuOptionToggleEnum>>> options) const;
```
jw098, line 38: Each MenuOptionItemEnum is paired with a vector MenuOptionToggleEnum. The reason for this is that the mapping for the options in each language is slightly different. e.g. For "Send to Boxes", the options may be Manual/Automatic in English, but may be On/Off in other languages. 

Here is some example code from AutoStory:

```
void AutoStory::change_settings(SingleSwitchProgramEnvironment& env, BotBaseContext& context, bool use_inference){
    env.console.log("Update settings.");
    if (use_inference){
        MenuOption session(env.console, context, LANGUAGE);

        std::vector<std::pair<MenuOptionItemEnum, std::vector<MenuOptionToggleEnum>>> options = {
            {MenuOptionItemEnum::TEXT_SPEED, {MenuOptionToggleEnum::FAST}},
            {MenuOptionItemEnum::SKIP_MOVE_LEARNING, {MenuOptionToggleEnum::ON}},
            {MenuOptionItemEnum::SEND_TO_BOXES, {MenuOptionToggleEnum::AUTOMATIC, MenuOptionToggleEnum::ON}},
            {MenuOptionItemEnum::GIVE_NICKNAMES, {MenuOptionToggleEnum::OFF}},
            {MenuOptionItemEnum::VERTICAL_CAMERA_CONTROLS, {MenuOptionToggleEnum::REGULAR, MenuOptionToggleEnum::NORMAL}},
            {MenuOptionItemEnum::HORIZONTAL_CAMERA_CONTROLS, {MenuOptionToggleEnum::REGULAR, MenuOptionToggleEnum::NORMAL}},
            {MenuOptionItemEnum::CAMERA_SUPPORT, {MenuOptionToggleEnum::ON}},
            {MenuOptionItemEnum::CAMERA_INTERPOLATION, {MenuOptionToggleEnum::NORMAL, MenuOptionToggleEnum::AVERAGE}},
            {MenuOptionItemEnum::CAMERA_DISTANCE, {MenuOptionToggleEnum::CLOSE}},
            {MenuOptionItemEnum::AUTOSAVE, {MenuOptionToggleEnum::OFF}},
            {MenuOptionItemEnum::SHOW_NICKNAMES, {MenuOptionToggleEnum::OFF, MenuOptionToggleEnum::DONT_SHOW}},
            {MenuOptionItemEnum::SKIP_CUTSCENES, {MenuOptionToggleEnum::ON}},
            {MenuOptionItemEnum::CONTROLLER_RUMBLE, {MenuOptionToggleEnum::ON}},
            {MenuOptionItemEnum::HELPING_FUNCTIONS, {MenuOptionToggleEnum::OFF}},

        };
        session.set_options(options);        
    }else{
        config_option(context, 1); // Text Speed: Fast
        config_option(context, 1); // Skip Move Learning: On
        config_option(context, 1); // Send to Boxes: Automatic
        config_option(context, 1); // Give Nicknames: Off
        config_option(context, 0); // Vertical Camera Controls: Regular
        config_option(context, 0); // Horiztontal Camera Controls: Regular
        config_option(context, 0); // Camera Support: On
        config_option(context, 0); // Camera Interpolation: Normal
        config_option(context, 0); // Camera Distance: Close
        config_option(context, 1); // Autosave: Off
        config_option(context, 1); // Show Nicknames: Don't show
        config_option(context, 1); // Skip Cutscenes: On
        config_option(context, 0); // Background Music: 10
        config_option(context, 0); // Sound Effects: 10
        config_option(context, 0); // Pokemon Cries: 10
        config_option(context, 0); // Controller Rumble: On
        config_option(context, 1); // Helping Functions: Off
    }

    pbf_mash_button(context, BUTTON_A, 1 * TICKS_PER_SECOND);
    clear_dialog(env.console, context, ClearDialogMode::STOP_TIMEOUT, 5);
    
}
```

SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_MenuOption.h
```diff
@@ -0,0 +1,83 @@
...
+
+private:
+
+    // set the current selected option, to the given option_toggle 
+    // (which is a list of option_toggles, since the mapping for the options in each language is slightly different)
+    void set_target_option(
+        const std::vector<MenuOptionToggleEnum> target_option_toggle_list
```
Mysticial, line 45: reference

SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_MenuOption.h
Mysticial: So in the case where there are multiple `MenuOptionToggleEnum`, it means it can be either one depending on the language. Thus the program to should set it to which ever one it finds?

jw098: Yes. The program keeps cycling the options, until it matches one of the option_toggles in the list.

SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_MenuOption.cpp
```diff
@@ -0,0 +1,214 @@
...
+    for (size_t c = 0; c < 10; c++){
+        m_boxes_item[c] = ImageFloatBox(0.055, 0.135 + c * 0.0739, 0.260, 0.060);
+        m_boxes_toggle[c] = ImageFloatBox(0.345, 0.135 + c * 0.0739, 0.175, 0.060);
+        m_overlays.add(COLOR_BLUE, m_boxes_item[c]);
+        m_overlays.add(COLOR_BLUE, m_boxes_toggle[c]);
+    }
+}
+
+void MenuOption::set_options(
+    std::vector<std::pair<MenuOptionItemEnum, std::vector<MenuOptionToggleEnum>>> options
```
Mysticial, line 50: I think there's several more places like this where you're copying a non-trivial parameter instead of passing by const-ref.

SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_MenuOption.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+    return read_option(cropped);
+}
+
+
+void MenuOption::move_to_option(const MenuOptionItemEnum target_option_enum) const{
+    std::string current_option_slug = read_option_item();
+    MenuOptionItem current_option = menu_option_item_lookup_by_slug(current_option_slug);
+
+    MenuOptionItem target_option = menu_option_item_lookup_by_enum(target_option_enum);
```
Mysticial, line 182: Both of these can be const-ref instead of value to avoid copy.

jw098: I can change target_option to const ref. But not current_option, since it gets reassigned in the loop.

SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_MenuOption.cpp
jw098: Basic programming question: For the vector of MenuOptionToggleEnum, within the pair, do I need to also make this a reference as well? Or is this redundant since I'm already making `options` pass by reference?

i.e.
`const std::vector<std::pair<MenuOptionItemEnum, std::vector<MenuOptionToggleEnum>&>>& options`
vs
`const std::vector<std::pair<MenuOptionItemEnum, std::vector<MenuOptionToggleEnum>>>& options`

Mysticial: The general rule of thumb for passing in read-only parameters is:

- If it's a primitive type (integer, float/double, pointer, enum): Pass by value. You don't need to mark it const because it's being copied and thus the function cannot affect the caller.
- If it's anything else, (struct, class, array, etc...): Pass by const reference. This avoids the copying the entire object since passing a reference is (internally), the same a passing a single pointer.

Of course exceptions will apply. Especially with async stuff where you have to watch out for the lifetimes of the objects.

PR #481, opened 2024-09-11
Update PokemonHome_BoxSorting.cpp

pifopi:


SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -99,14 +99,14 @@ BoxSorting::BoxSorting()
...
           LockMode::LOCK_WHILE_RUNNING,
-          20
+          50
           )
     , GAME_DELAY(
-          "<b>Delay of your Pokemon Home app (default value should be fine):</b>",
+          "<b>Delay of your Pokemon Home app:</b>",
           LockMode::LOCK_WHILE_RUNNING,
-          15
+          30
```
pifopi, line 109: I went the safe way because many people reoprted issues so trading longer running time for easier to reach reliability is better

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -618,14 +617,10 @@ void BoxSorting::program(SingleSwitchProgramEnvironment& env, BotBaseContext& co
...
-                            extract_box_reference(screen, national_dex_number_box),
-                            0xff808080, 0xffffffff, true
-                            );
-
-                        int national_dex_number = OCR::read_number(env.console, image);
-                        if (national_dex_number == -1){
-                            dump_image(env.console, ProgramInfo(), "ReadSummary", screen);
+
+                        int national_dex_number = OCR::read_number_waterfill(env.console, extract_box_reference(screen, national_dex_number_box), 0xff808080, 0xffffffff);
+                        if (national_dex_number < 0 || national_dex_number > 1025) {
```
pifopi, line 622: it will prevent this to run with next game but we have time and it might not even be on the switch. But that way, we can catch misread for current pokemon

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorting.cpp
```diff
@@ -647,14 +642,9 @@ void BoxSorting::program(SingleSwitchProgramEnvironment& env, BotBaseContext& co
...
-                        image = to_blackwhite_rgb32_range(
-                            extract_box_reference(screen, ot_id_box),
-                            0xff808080, 0xffffffff, true
-                            );
-
-                        int ot_id = OCR::read_number(env.console, image);
-                        if (ot_id == -1){
-                            dump_image(env.console, ProgramInfo(), "ReadSummary", screen);
+                        int ot_id = OCR::read_number_waterfill(env.console, extract_box_reference(screen, ot_id_box), 0xff808080, 0xffffffff);
+                        if (ot_id < 0 || ot_id > 999'999) {
```
pifopi, line 646: same here, i think that is the maximum value for OT

PR #484, opened 2024-09-13
Autostory tools

jw098:
Should be merged after #482.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStoryTools.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStoryTools.cpp
@@ -1,0 +1,666 @@
+/*  AutoStoryTools
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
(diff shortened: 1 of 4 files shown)


SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStoryTools.cpp
```diff
@@ -0,0 +1,666 @@
...
+            console, context,
+            std::chrono::seconds(90),
+            {battle, overworld, dialog}
+        );
+        context.wait_for(std::chrono::milliseconds(100));
+
+        switch (ret){
+        case 0: // battle
+            console.log("Detected battle menu, spam first move.");
+            pbf_mash_button(context, BUTTON_A, 3 * TICKS_PER_SECOND);
```
Mysticial, line 71: I assume that it should be rare that we manage to mash into the move selection, but not past it right? IIRC, we do have a detector for the move selection in case we happen to get stuck there.

jw098: As far as I know, it shouldn't be possible to mash past the move selection, since the move animation takes so long.

PR #485, opened 2024-09-14
Autostory base

jw098:
The Autostory.h file, on its own.
#484 should be merged first.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStory.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStory.cpp
@@ -15,12 +15,22 @@
 #include "Pokemon/Pokemon_Strings.h"
 #include "PokemonSwSh/Inference/PokemonSwSh_IvJudgeReader.h"
 #include "PokemonSV/Programs/PokemonSV_GameEntry.h"
-// #include "PokemonSV_AutoStory_Checkpoint_00_04.h"
-// #include "PokemonSV_AutoStory_Checkpoint_05_10.h"
-// #include "PokemonSV_AutoStory_Checkpoint_11_15.h"
-// #include "PokemonSV_AutoStory_Checkpoint_16_20.h"
-// #include "PokemonSV_AutoStory_Checkpoint_21_25.h"
-// #include "PokemonSV_AutoStory_Checkpoint_26_30.h"
+#include "PokemonSV_AutoStory_Segment_00.h"
+// #include "PokemonSV_AutoStory_Segment_01.h"
+// #include "PokemonSV_AutoStory_Segment_02.h"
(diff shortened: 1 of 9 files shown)


SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStory.cpp
```diff
@@ -0,0 +1,648 @@
...
+    // checkpoint_list.push_back([&](){checkpoint_34(env, context, notif_status_update);});
+
+    for (int checkpoint = start; checkpoint <= end; checkpoint++){
+        if (checkpoint == 0){
+            console.log("checkpoint_0");
+            checkpoint_list[checkpoint]();
+            continue;
+        }
+        
+        std::string leading_zero = checkpoint < 10 ? "0" : "";
```
Mysticial, line 428: A cleaner way to do this is:
```
const size_t DIGITS = 3;
std::string number = std::to_string(checkpoint);
if (number.size() < DIGITS){
    number = std::string(DIGITS - number.size(), '0') + number;
}
```
which will work for any # of digits.

SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStory.cpp
```diff
@@ -0,0 +1,648 @@
...
+    case StartPoint::LOS_PLATOS:
+        return "Start: Battled Arven, received Legendary's Pokeball. Talked to Nemona at Lighthouse.";
+    case StartPoint::MESAGOZA_SOUTH:
+        return "Start: At Los Platos Pokecenter.";
+    default:
+        return "";        
+    }
+}
+
+std::string AutoStory::end_segment_description(){
```
Mysticial, line 350: There's one of these for every segment right? How many segments are we expecting? Like 50+ of them?

I'm wondering if this should be encapsulated into the the segments themselves. Like we define an abstract class representing a segment, then we have each segment inherit it and implement all the required stuff such as these start/end messages and the actual logic.

This approach will require refactoring the segment dropdown away from EnumDropdown to something more generic like the StringSelect options which we use in places like Pokemon name dropdowns. These are built from arrays rather than compile-time lists of enums.

jw098:
I refactored Autostory so that each segment is it's own class. This let's us run the segments from the list, based on index, instead of a large switch statement.

We want our autostory to have a break after the tutorial, to allow the user to send in the level 100 Gardevoir as the lead. This is why I added the EnumDropdownOption STORY_SECTION, to toggle between the Tutorial and Main story segments. But since we're trying to PR the code in stages, the main story segments haven't been PR'ed yet. This is why StringSelectOption STARTPOINT_MAINSTORY and ENDPOINT_MAINSTORY are commented out.

PR #487, opened 2024-09-18
add North detection to Overworld watcher

jw098:


SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_DirectionDetector.cpp
```diff
@@ -77,7 +88,7 @@ std::pair<double, double> DirectionDetector::locate_north(ConsoleHandle& console
     if (zero_coord.first > 0.91 || zero_coord.first < 0.90 || 
         zero_coord.second > 0.84 || zero_coord.second < 0.82)
     {
-        console.log("Unable to locate the overworld radar ball, falling back on hard coded location.", COLOR_ORANGE);
+        std::cout << "Unable to locate the overworld radar ball, falling back on hard coded location." << std::endl;
```
Mysticial, line 91: Stick with `console.log()` for normal logging and reserve `std::cout` only for debugging.

jw098: If we use console in locate_north, it means that OverworldWatcher can no longer use it (via detect_north), since OverworldWatcher doesn't contain console.

So, I see 2 other options:
- Refactor OverworldWatcher to contain console in its constructor.
- Leave console within locate_north. But then copy its code into detect_north, and remove the sections that require console. This allows OverworldWatcher to still use detect_north. But this creates code duplication.

PR #500, opened 2024-11-02
NoMinimapDetector

jw098:
Detects when Minimap no longer visible.
Useful for detecting when you finished the Olive roll.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_NoMinimapDetector.cpp
+++ b/SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_NoMinimapDetector.cpp
@@ -1,0 +1,44 @@
+/*  No Minimap Detector
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+
+#include "Kernels/Waterfill/Kernels_Waterfill_Session.h"
+#include "CommonFramework/Globals.h"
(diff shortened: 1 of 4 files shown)


SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_NoMinimapDetector.cpp
```diff
@@ -0,0 +1,44 @@
...
+//using std::endl;
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonSV{
+
+
+
+NoMinimapDetector::NoMinimapDetector(Logger& logger, Color color)
+    : m_color(color)
```
Mysticial, line 28: Why do we need `m_color` and `m_ball` separately just for the overlay? Can't we make `m_overworld` do the overlay directly?

jw098: I wanted NoMinimapDetector's overlay to look slightly different than the Overworld detector's overlay. Overworld detector has 3 boxes in its overlay, while NoMinimapDetector's overlay is only the smallest inner box. This way, users can tell when we're looking or the Overworld, and when we're looking for no overworld.

I don't feel strongly about this though. If you prefer to use the Overworld's overlay, I can change it.

PR #502, opened 2024-11-03
OliveDetector

jw098:
Olive detector and all the associated functions for pushing the olive.

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_OliveDetector.cpp
+++ b/SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_OliveDetector.cpp
@@ -1,0 +1,386 @@
+/*  Olive Detector
+ *
+ *  From: https://github.com/PokemonAutomation/Arduino-Source
+ *
+ */
+#include "Kernels/Waterfill/Kernels_Waterfill_Session.h"
+#include "CommonFramework/Globals.h"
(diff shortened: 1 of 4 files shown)


SerialPrograms/Source/PokemonSV/Inference/Overworld/PokemonSV_OliveDetector.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonSV{
+
+
+class OliveMatcher : public ImageMatch::WaterfillTemplateMatcher{
+public:
+    OliveMatcher() : WaterfillTemplateMatcher(
+        "PokemonSV/Olive.png", Color(0,0,0), Color(255, 255, 255), 5
```
Mysticial, line 35: How is this supposed to work? The olive can in be multiple orientations as well as being eclipsed or occulted.

jw098: All non-green pixels are filtered out. Then the largest blob of green is considered to be the Olive. Unlike all the other detectors, the RMSD doesn't matter much. This is why it doesn't care if the olive is eclipsed or in a different orientation, as long as it's the largest blob of green.

Mysticial:
Rebase

PR #505, opened 2024-11-10
MacOS QT Permissions Fix

Ericzklm:
Needed to re-install the program on the latest MacOS version; thought I would see if there was a way to fix the current issue on MacOS ARM.

In order for MacOS to prompt the user for camera and microphone permissions, the qt target needs to be finalized after all libraries have been linked. This addition allows for permission requests to be sent with the latest version of QT on MacOS. I don't have the means to test this change on Windows.

MacOS Version: Sequoia 15.1
Xcode 16.1 (16B40)
Apple clang version 16.0.0 (clang-1600.0.26.4)
Homebrew 4.4.4
brew list --versions:
- cmake 3.31.0
- qt 6.7.2_2
- tesseract 5.4.1_2
- opencv 4.10.0_12

### DIFF
--- a/SerialPrograms/CMakeLists.txt
+++ b/SerialPrograms/CMakeLists.txt
@@ -2056,8 +2056,8 @@ else() # WIN and Linux:
 endif()
 
 set_target_properties(SerialPrograms PROPERTIES LINKER_LANGUAGE CXX)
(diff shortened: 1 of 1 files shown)


Gin890:
Thank you!

PR #507, opened 2024-11-14
Mac Bundle Fixes

Ericzklm:
Various fixes relating to the MacOS SerialPrograms app bundle

- Modifying external filepaths to use the .app base directory instead of the executable's Contents/MacOS/ directory
  - Needed in order to enable SerialPrograms.log and other dumps. This stops the program from attempting to write in write-protected space inside the bundle
  - Needed to avoid collision between the SerialProgram's "Resources" folder and the "Resources" folder automatically defined in the .app bundle
- Modified the location of the SerialPrograms.log to use an absolute filepath located outside the .app bundle for MacOS
- Created an Apple Icon Image file and linked it to the .app bundle during build


SerialPrograms/CMakeLists.txt
```diff
@@ -2034,7 +2034,7 @@ source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR}../../ FILES ${MAIN_SOURCES})
 
 if (APPLE)
     # TODO: set up the correct application icon
-    # set(SerialPrograms_ICON ${CMAKE_CURRENT_SOURCE_DIR}/../Resources/icon.icns)
+    set(SerialPrograms_ICON ${CMAKE_CURRENT_SOURCE_DIR}/../IconResource/icon.icns)
```
Mysticial, line 2037: Can we pick a different name than "IconResources"? Since there's a lot more than just icons.

Gin890:
Thank you for the PR. I tried to add an icon image to the program but for some reason I could not get it working. The compiled program did not show the icon correctly. Hope this PR solves that!

Ericzklm:
Sorry didn't get a chance to follow-up before the merge. 

The IconResource folder looked to contain only icon related files so I thought to stick the icon.icns in there as well. It can be moved into the SerialPrograms Resources if that's a better location for it.

The Icon should work! I was confused for a bit because the Icon wasn't showing up when it should have but the problem was just that Finder had cached the blank icon.

PR #508, opened 2024-11-15
Check QT version

jw098:


Gin890:
Thank you for the fix! It should be easy to fix for QT 6.1 in a follow-up PR. Let me accept the PR first to fix the CI

PR #510, opened 2024-11-16
Segment 11 (Olive roll) and update AutoStoryTools

jw098:


SerialPrograms/Source/PokemonSV/Programs/AutoStory/PokemonSV_AutoStoryTools.cpp
```diff
@@ -1065,6 +1090,76 @@ void move_cursor_towards_flypoint_and_go_there(
 }
 
 
+
+void change_date(
```
Mysticial, line 1094: This function looked familiar. Then I realized the ItemPrinterRNG does something different since it needs to keep track of when it commits.

Let's move this function into somewhere in NintendoSwitch/ since I feel like a lot of things can use it.

jw098:
FYI, the reason I need to change the date prior to the Olive roll is that Olive detection is dependent on certain lighting conditions. Night time gives the most consistent results for the Olive roll.

PR #511, opened 2024-11-16
Segments 12 to 13

jw098:


SerialPrograms/Source/PokemonSV/Programs/PokemonSV_Navigation.cpp
```diff
@@ -553,7 +553,7 @@ void leave_phone_to_overworld(const ProgramInfo& info, ConsoleHandle& console, B
             ErrorReport::SEND_ERROR_REPORT, console,
             "leave_phone_to_overworld(): Unknown state after 10 button Y presses.",
             true
-        );
+        );
```
Mysticial, line 556: Leftover from temporary code?

PR #525, opened 2025-01-15
ruby/sapphire starter reset

kichithewolf:
I plugged the microcontroller into a retro handheld (RG35XXSP, stock OS) and it worked, so I wrote this because new toy. Probably going to aim for a legendary reset (primarily Deoxys since that's high value) or do starter reset for FRLG next. In the future, assuming we ever get GBA Pokemon on the Switch or Switch 2, it should be easy to modify to work there.

As far as I can tell, the video output on the handheld I have matches that of the Switch's GBA emulator. I don't have NSO+, but managed to find an album ([here](https://imgur.com/a/nso-gb-gba-display-options-F3WfjNW)) of the Switch's display options for its emulator, which matches what I have when set to Full Screen, no filter.

In theory, this could probably run on cart using an analogue pocket w/ dock, if the microcontroller worked (and the user turns off any audio enhancements). Searching discord says no, but I'm curious if a USB controller adapter such as the Magic-NS would make it work. But I'm not willing to spend $350+ to find out.

---

The RSE games are locked behind developer mode in the dropdown, under the assumption that anyone who can figure out how to enable developer mode can also figure out any issues they run into due to differing devices or whatever.

FeedbackType VIDEO_AUDIO_GBA was added to avoid the regular black border check. BlackBorderGBADetector was created to go with it. It is exactly the same as black border detector, but with the GBA's borders. The original BBD should work with GB and GBC when set to full screen, but the GBA's aspect ratio means that it kept setting off the check.

The sound listener program is a straight copy of the one from PLA/BDSP since I needed it for testing. Same with ShinySoundDetector, it's the same as existing sound detection.

The version of starter reset under developer mode (StarterReset, not AudioStarterReset) was created because I originally couldn't get sound working. It still works by checking the starter's summary screen in battle. If the starter is shiny, its pokedex number in the summary will be gold-ish instead of white. Its redundant and slower, so I hid it in favor of AudioStarterReset.

---

Note about the shiny sound: I originally tried using a rip of the sound effect, but couldn't get it working. I then recorded myself sending out a shiny on an emulator on my PC (a different emulator from the one I'm running on the handheld) and just trimmed the audio in Audacity. I tried to isolate the sound channels on the PC emulator, but that didn't work, the shiny sound plays with the majority of the background music. As a result it's not really the best clip. Except as far as I can tell the high quality rip was too good, and the emulators are faithfully terrible? Either way it's probably going to need replacement at some point.


SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
```diff
@@ -42,6 +42,7 @@ const std::set<std::string> TOKENS{
     "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08", //  jw's token.
     "e8d168bc482e96553ea9f9ecaea5a817474dbccc2a6a228a6bde67f2b2aa2889", //  James' token.
     "7555b7c63481cad42306718c67e7f9def5bfd1da8f6cd299ccd3d7dc95f307ae", //  Kuro's token.
+    "3d475b46d121fc24559d100de2426feaa53cd6578aac2817c4857a610ccde2dd", //  kichi's token.
```
Mysticial, line 45: Feel free to pull this out into a separate PR so it can go in earlier.

kichithewolf:
Closed PR as the last of it went in with https://github.com/PokemonAutomation/Arduino-Source/pull/530.

PR #527, opened 2025-01-26
rse new panel and settings

kichithewolf:
split out from https://github.com/PokemonAutomation/Arduino-Source/pull/525

only the new game panel/setting setup


SerialPrograms/Source/PokemonRSE/PokemonRSE_Settings.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+
+GameSettings& GameSettings::instance(){
+    static GameSettings settings;
+    return settings;
+}
+GameSettings::GameSettings()
+    : BatchOption(LockMode::LOCK_WHILE_RUNNING)
+    , m_menu_navigation("<font size=4><b>Menu Navigation Timings:</b></font>")
+    , GAME_TO_HOME_DELAY(
```
Mysticial, line 25: How many of these are relevant to GBA? Stuff like the Home menu didn't exist on the real thing.

PR #532, opened 2025-02-01
Mac qt6.5 fix

Ericzklm:
Adding a workaround on MacOS to bypass issue of undetected video feed
Screenshot 2025-02-01 at 3 03 49 PM

All frames returned from the video sink on MacOS have an invalid presentation time despite everything else being valid. Not clear on why, possibly a QT bug or there's some mac-specific configuration that's missing.

The program thinks the video feed is not working when it actually is due to this and removing the check for duplicate frames gets around this. Not the best solution but the workaround is functional.


SerialPrograms/Source/CommonFramework/VideoPipeline/Backends/CameraWidgetQt6.5.cpp
```diff
@@ -275,9 +275,11 @@ void CameraSession::connect_video_sink(QVideoSink* sink){
                 WriteSpinLock lg(m_frame_lock);
 
                 //  Skip duplicate frames.
+                #ifndef __APPLE__
```
Mysticial, line 278: Better idea: Skip the check is the timestamp is invalid instead of if-def'ing on the platform.

PR #548, opened 2025-03-11
LGPE: game resets

kichithewolf:
Added in resetting game from home, based off the resets from the other games. Does not yet handle the right joycon update, update can be suppressed for an unknown amount of time by going to home from the game and declining the update manually.

One thing of note - starting LGPE seems to disconnect all controllers. The ESP32 reconnects and continues to enter the game, but one time it didn't. It still showed as connected and paired, and the logs showed the inputs I attempted, but game/console was non-responsive. I haven't been able to replicate but its something to keep an eye out for.

Also, it works:
image


SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_GameEntry.cpp
```diff
@@ -312,6 +312,206 @@ bool openedgame_to_gamemenu(
...
+    }
+
+    if (game_slot != 0){
+        ssf_press_button(context, BUTTON_HOME, ConsoleSettings::instance().SETTINGS_TO_HOME_DELAY0, 160ms);
+        for (uint8_t c = 1; c < game_slot; c++){
+            pbf_move_joystick(context, 255, 128, 80ms, 0ms);
+        }
+    }
+
+    if (tolerate_update_menu){
```
Mysticial, line 442: I think we can simplify this logic now and assume that feedback is available. This fallback system was for the old SwSh programs that could run without feedback.

PR #558, opened 2025-03-29
update read_number_waterfill(), Fix Itemprinter material detection

jw098:


SerialPrograms/Source/CommonTools/OCR/OCR_NumberReader.cpp
```diff
@@ -104,52 +105,94 @@ int read_number_waterfill(
...
+                ocr_text += ocr[0];
+            }else{
+                empty_char = true;
+                break;
+            }
         }
-    }
+        if (empty_char){
+            // try the next color filter
+            continue;
```
Mysticial, line 179: Rather than breaking out on the first match, you might want to run all of them anyway, then do a majority vote on the results (after throwing out obvious incorrect reads like empty string or letters). TeraCardReader does this with the star count: https://github.com/PokemonAutomation/Arduino-Source/blob/ba9d5bdc7254387bc601dfde8675e20250f6b9e7/SerialPrograms/Source/PokemonSV/Inference/Tera/PokemonSV_TeraCardDetector.cpp#L132

The reason is that on the borderline cases may return neither an empty string nor a valid result. You'll want to result when the "signal is strongest".

PR #585, opened 2025-05-05
fixes building on arm

naussika:


SerialPrograms/Source/Kernels/ImageFilters/RGB32_Range/Kernels_ImageFilter_RGB32_Range.h
```diff
@@ -32,7 +32,7 @@ size_t filter_rgb32_range(
 //  The purpose is to reduce passes over the entire image.
 //  All matricies must have the same dimensions.
 struct FilterRgb32RangeFilter{
-    uint32_t* const data;       //  Pointer will be overwritten.
+    uint32_t* data;       //  Pointer will be overwritten.
```
Mysticial, line 35: Does this `const` actually break the build on Mac? I added it since my refactor makes it so it no longer changes this pointer value. (the comment is out of date)

naussika: with `const` the error would be:
```
Kernels_ImageFilter_RGB32_Range_Routines.h:49:33: error: cannot assign to non-static data member 'data' with const-qualified type 'uint32_t *const' (aka 'unsigned int *const')
   49 |                     filter.data = (uint32_t*)((const char*)filter.data + filter.bytes_per_row);
      |                     ~~~~~~~~~~~ ^
Kernels_ImageFilter_RGB32_Range.h:35:21: note: non-static data member 'data' declared const here
   35 |     uint32_t* const data;       //  Pointer will be overwritten.
      |     ~~~~~~~~~~~~~~~~^~~~
```

Mysticial:
I'll merge, then I'll add back the const and it should still build.

PR #600, opened 2025-06-06
Home to date time

jw098:
Using inference for better reliability.

Right now, I created a second version of `home_to_date_time`, which includes VideoStream. If you think it looks good, we can delete the old `home_to_date_time` and update all the references.'

### DIFF
--- a/SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_Navigation.cpp
+++ b/SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_Navigation.cpp
@@ -35,29 +35,40 @@ bool is_setting_selected(VideoStream& stream, ProControllerContext& context, Ima
     overlays.add(COLOR_BLUE, unselected_box2);
     context.wait_for_all_requests();
     context.wait_for(Milliseconds(250));
-    VideoSnapshot snapshot = stream.video().snapshot();
+    size_t max_attempts = 5;  // multiple attempts because the highlighted icon/setting pulses. and sometimes there isn't enough contrast at the exact moment you take the snapshot.
+    bool is_selected = false;
+    for (size_t i = 0; i < max_attempts; i++){
+        VideoSnapshot snapshot = stream.video().snapshot();
 
-    ImageStats stats_unselected_box1 = image_stats(extract_box_reference(snapshot, unselected_box1));
-    double unselected1_average_sum = stats_unselected_box1.average.sum();
-    stream.log("unselected_average_sum1: " + std::to_string(unselected1_average_sum));
+        ImageStats stats_unselected_box1 = image_stats(extract_box_reference(snapshot, unselected_box1));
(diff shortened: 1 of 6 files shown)


SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_Navigation.cpp
```diff
@@ -5,14 +5,318 @@
...
+        ssf_issue_scroll(context, SSF_SCROLL_DOWN, 3);
+        ssf_issue_scroll(context, SSF_SCROLL_DOWN, 4);
+
+        ssf_issue_scroll(context, SSF_SCROLL_LEFT, 0);
+
+        // we expect to be at Settings icon
+        ImageFloatBox system_settings(0.685, 0.69, 0.05, 0.03);
+        ImageFloatBox other_setting1(0.615, 0.69, 0.05, 0.03);
+        ImageFloatBox other_setting2(0.545, 0.69, 0.05, 0.03);
+        if (!is_setting_selected(stream, context, system_settings, other_setting1, other_setting2)){
```
Mysticial, line 91: This sounds like it might slow it down quite a bit? An alternative that I'm thinking about is to keep the A press to enter the settings, then run `run_until()` with the down-scroll mash until you detect that we're at the bottom of the settings menu.

SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_Navigation.cpp
```diff
@@ -5,14 +5,261 @@
...
+        ssf_issue_scroll(context, SSF_SCROLL_DOWN, 3);
+        // }
+
+
+        //  double up this A press to make sure it gets through.
+        ssf_press_button(context, BUTTON_A, 3);
+        ssf_press_button(context, BUTTON_A, 3);
+
+        context.wait_for_all_requests();
+        context.wait_for(Milliseconds(250));
```
jw098, line 105: For some reason, `wait_for_all_requests()` alone doesn't actually wait for all the buttons to finish pressing. This is why I have the extra 250ms wait here. Any ideas on why this might be the case?

Mysticial: That's not supposed to happen. Needing a delay before inference is rather common since the video + Switch's UI delay is longer than the time to process return trip for a finished command.

Is the following `wait_until()` triggering early without this wait?

jw098: The program still works as intended without the wait.
But I can see that the wait_until inference is starting a bit earlier than I want. I want the button presses to be completely done before we start inference, to reduce the probability of early false positives. However, this is a theoretical concern, since I haven't seen any issues in my testing, even without the 250ms wait.

PR #603, opened 2025-06-09
BoxDraw: add custom text box for ImageFloatBox coordinates.

jw098:


SerialPrograms/Source/NintendoSwitch/DevPrograms/BoxDraw.cpp
```diff
@@ -57,16 +59,25 @@ class BoxDraw::DrawnBox : public ConfigOption::Listener, public VideoOverlay::Mo
...
+            m_parent.BOX_COORDINATES.add_listener(*this);
             overlay.add_listener(*this);
         }catch (...){
             detach();
             throw;
         }
     }
     virtual void on_config_value_changed(void* object) override{
-        std::lock_guard<std::mutex> lg(m_lock);
+        // std::lock_guard<std::mutex> lg(m_lock);
```
jw098, line 70: Within on_`config_value_changed()`, is the mutex lock guard necessary? 

Without the lock guard, the program works as intended. With the lock guard, the program freezes.
With the lock guard, I feel that a bidirectional association between two config values within the same program is very difficult without doing some weird tricks with the locks. 

e.g. see here: https://github.com/PokemonAutomation/Arduino-Source/blob/main/SerialPrograms/Source/PokemonSV/Programs/ItemPrinter/PokemonSV_ItemPrinterRNGTable.cpp#L96

The Config.set() functions already have a spinlock on them. And they already check that the value is actually changing, so we don't get an infinite loop. Does this program require the mutex to be thread safe?

Mysticial: In this particular case the lock is not necessary because all changes to the configs are from the UI and thus the same thread.

The general case where the program thread can also change the config is where a lock is needed. Thus is became standard practice to slap a lock over all internal state.

Does replacing it with a recursive mutex work? (yes this is super ugly) Otherwise, I'm fine with removing the mutex as long as this `BoxDraw` class remains local this this program and thus is only modified from the main UI thread.

PR #630, opened 2025-06-27
Save size, position for Main Window, Output Window on program start

jw098:


jw098:
After pushing the initial PR, I realized there were a few other features I wanted to implement.
Todo:
ensure INITIAL_X_POS can’t be too negative. Set minimum value based on window width/height.
Add event listener so that changing INITIAL_X_POS in the config will also change the window position.

SerialPrograms/Source/CommonFramework/Windows/MainWindow.cpp
```diff
@@ -46,6 +47,16 @@ MainWindow::MainWindow(QWidget* parent)
         GlobalSettings::instance().WINDOW_SIZE->WIDTH,
         GlobalSettings::instance().WINDOW_SIZE->HEIGHT
     );
+    // move main window to desired position on startup
+    auto const screen_geometry = QGuiApplication::primaryScreen()->availableGeometry();
+    uint32_t const screen_width = (uint32_t)screen_geometry.width();
+    uint32_t const screen_height = (uint32_t)screen_geometry.height();
+    uint32_t initial_x_pos_main = GlobalSettings::instance().WINDOW_SIZE->INITIAL_X_POS;
+    uint32_t initial_y_pos_main = GlobalSettings::instance().WINDOW_SIZE->INITIAL_Y_POS;
+    uint32_t move_x_main = std::min(initial_x_pos_main, screen_width-100);
```
Mysticial, line 56: What happens if this goes negative? `screen_width-100 < 0`

jw098: I assumed that PA would be run on screens bigger than 100 pixels. But fair enough, I'll change it to `screen_width*0.97`.

Issue #629, opened 2025-06-26
Remove sleepy dependencies and associated code


pifopi:
Sleepy was depreacated for a while and could now be deleted

PR #631, opened 2025-06-27
Remove sleepy

pifopi:
Will fix https://github.com/PokemonAutomation/Arduino-Source/issues/629


pifopi:
Ready for when v0.54 goes completely live to public release

PR #632, opened 2025-06-29
Autostory fixes

jw098:


Mysticial:
This may be conflicting with the line ending change. Change all the line endings to CRLF and it should resolve.

PR #638, opened 2025-06-30
gitattributes: disable all line ending conversion

jw098:


pifopi:
Can you execute

git add --renormalize .

It did all that changes on my side,
image
Those files are currently only having LF and based on what I understood from Kuro, he'd want them CRLF (maybe not the .rc file ?)

jw098:
I previously ran `git add --renormalize .`, and I got the same result as you. But when I run it again, nothing happens.

I'm not sure what's happening.

Mysticial:
Can we confirm that this still allows merges and diffs like normal? If so, we're good to merge.

PR #674, opened 2025-08-26
Fixes memory stats on macOS

naussika:


Gin890:
The code does not match what I saw on my mac. The code computes total used memory as 32 GB but on my mac it's actually 40 GB. the process used memory is also larger than it should be. Activity Monitor says it's 319 MB but the code computes 450MB.

Gin890:
LGTM! Sorry I didn't check the update of this PR sooner!

PR #676, opened 2025-08-30
Implement Discord Social SDK

Koi-3088:
Allows setting rich presence for the user running ComputerControl. Displays program name, status, elapsed time.

Enabled by default, can be toggled off. Requires a Discord client to be running on user's computer.

Includes debug and release .lib and binary files for Windows, Linux, MacOS. Linux and MacOS building currently not implemented.

### DIFF
--- a/SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
+++ b/SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
@@ -1,0 +1,170 @@
+#ifdef PA_SOCIAL_SDK
+#define DISCORDPP_IMPLEMENTATION
+
+#include <thread>
+#include "DiscordSocial.h"
+#include "CommonFramework/Logging/Logger.h"
+#include "Integrations/ProgramTracker.h"
+#include "../../CommonFramework/GlobalSettingsPanel.h"
+
+using namespace discordpp;
+namespace PokemonAutomation{
+namespace Integration{
+namespace DiscordSocialSDK{
+
+DiscordSocial& DiscordSocial::instance(){
+    static DiscordSocial instance;
+    return instance;
+}
+
+void DiscordSocial::run(){
+    m_client = std::make_shared<Client>();
+    if (!m_client){
+        log("Failed to initialize DiscordSocialSDK.", "run()", LoggingSeverity::Error);
+        m_client.reset();
+        return;
+    }
+
+    m_client->SetApplicationId(m_app_id);
+    m_client->AddLogCallback([&](auto message, auto severity){
+        log(message, "Internal", severity);
+    }, m_log_level);
+
+    std::thread(&DiscordSocial::thread_loop, this).detach();
+}
+
+void DiscordSocial::thread_loop(){
+    log("Starting Discord Rich Presence update thread...", "thread_loop()", LoggingSeverity::Info);
+    m_running.store(true, std::memory_order_release);
+    m_activity = create_activity();
+    m_client->UpdateRichPresence(m_activity, [&](ClientResult result){
+        if (!result.Successful()){
+            log("Rich Presence initial update failed.", "thread_loop()", LoggingSeverity::Warning);
+        }
+    });
+
+    while (m_running.load(std::memory_order_relaxed)){
+        try{
+            if (!m_running.load(std::memory_order_relaxed) || !GlobalSettings::instance().RICH_PRESENCE){
+                break;
+            }
+
+            RunCallbacks();
+            update_rich_presence();
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+        }catch (const std::exception& e){
+            log("Exception: " + std::string(e.what()), "thread_loop()", LoggingSeverity::Error);
+            break;
+        }
+    }
+
+    log("Discord Rich Presence update thread exiting...", "thread_loop()", LoggingSeverity::Info);
+    m_running.store(false, std::memory_order_release);
+    m_client.reset();
+}
+
+void DiscordSocial::update_rich_presence(){
+    try{
+        std::string details = m_activity.Details().value();
+        std::string state = m_activity.State().value();
+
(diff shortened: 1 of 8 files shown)


Koi-3088:
I guess it would also close [this](https://github.com/PokemonAutomation/ComputerControl/issues/109)

3rdPartyBinaries/discord_social_sdk.zip
Mysticial, line 1: Can you split this into the files for each OS? Then only unzip for that OS that's being used.

SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.h
```diff
@@ -0,0 +1,41 @@
...
+            Logger& logger();
+            void log(const std::string& message, const std::string& identity, const discordpp::LoggingSeverity& severity);
+            discordpp::Activity create_activity();
+            void update_rich_presence();
+            void thread_loop();
+
+        private:
+            std::shared_ptr<discordpp::Client> m_client = nullptr;
+            std::atomic<bool> m_running;
+            const uint64_t m_app_id = 1406867596585865326;
```
Mysticial, line 33: Is this intentionally hard-coded? Where does it come from?

Koi-3088: `m_app_id` is intentionally hardcoded. It's the private Discord application the SDK gets art assets from.

SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
```diff
@@ -0,0 +1,170 @@
...
+    activity.SetState("Idle");
+
+    ActivityButton button1{};
+    button1.SetLabel("Download from GitHub");
+    button1.SetUrl("https://github.com/PokemonAutomation/ComputerControl/releases/latest");
+    activity.AddButton(button1);
+
+    ActivityButton button2{};
+    button2.SetLabel("Join our Discord");
+    button2.SetUrl("https://discord.gg/pokemonautomation");
```
Mysticial, line 154: You can use the URLs defined here: https://github.com/PokemonAutomation/Arduino-Source/blob/main/SerialPrograms/Source/CommonFramework/Globals.cpp

SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
```diff
@@ -0,0 +1,170 @@
...
+    activity.AddButton(button2);
+
+    discordpp::ActivityAssets assets{};
+    assets.SetLargeImage("rotom");
+    assets.SetLargeText("Pok\u00E9mon Automation");
```
Mysticial, line 159: More strings you can pull here: https://github.com/PokemonAutomation/Arduino-Source/blob/main/SerialPrograms/Source/CommonFramework/Globals.cpp#L46

SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
```diff
@@ -0,0 +1,171 @@
...
+        m_client.reset();
+        return;
+    }
+
+    m_client->SetApplicationId(m_app_id);
+    m_client->AddLogCallback([&](auto message, auto severity){
+        log(message, "Internal", severity);
+    }, m_log_level);
+
+    std::thread(&DiscordSocial::thread_loop, this).detach();
```
Mysticial, line 34: Never use detach since it's impossible to clean up due to having an unbounded lifetime. If you destruct this class, and the thread is still running, you dangle all the fields that it references.

Store the thread into the class. Then in the destructor after you signaled the thread to exit, join the thread.
```
m_thread = std::thread(&DiscordSocial::thread_loop, this);
```

In destructor:
```
m_running.store(false, std::memory_order_release);
m_thread.join();

//  Now it's safe to destroy the client.
if (m_client) m_client.reset();
```

SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
```diff
@@ -0,0 +1,171 @@
...
+DiscordSocial& DiscordSocial::instance(){
+    static DiscordSocial instance;
+    return instance;
+}
+
+void DiscordSocial::run(){
+    m_client = std::make_shared<Client>();
+    if (!m_client){
+        log("Failed to initialize DiscordSocialSDK.", "run()", LoggingSeverity::Error);
+        m_client.reset();
```
Mysticial, line 25: This isn't exception safe. If an exception is thrown further down, `m_client` remains initialized which is probably not what you intended.

Make the client first and store into a local variable:
```
std::shared_ptrClient> client = std::make_shared<Client>();
```
Then at the the end of the function (before you start the thread), move it into `m_client`. 

It's possible for the thread creation itself to throw. So to be truly correct, you'll need something like this:

```
m_client = std::move(client);
try{
    std::thread(&DiscordSocial::thread_loop, this).detach();
}catch (...){
    m_client.reset();
    throw;
}
```
But see my other comment about not using `detach()`.

SerialPrograms/Source/Integrations/DiscordSocial/DiscordSocial.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+    while (m_running.load(std::memory_order_relaxed)){
+        try{
+            if (!m_running.load(std::memory_order_relaxed) || !GlobalSettings::instance().RICH_PRESENCE){
+                break;
+            }
+
+            RunCallbacks();
+            update_rich_presence();
+            std::this_thread::sleep_for(std::chrono::seconds(1));
```
Mysticial, line 55: Do we need to do this every second? Is it possible to diff to see if anything changed. And only update if something changed?

Ideally CC itself should push these changes out, but now is not the time to re-architect that part so this is fine.

Mysticial (approved):

PR #690, opened 2025-09-12
Record Keyboard Controller

jw098:


SerialPrograms/Source/Controllers/KeyboardInput/KeyboardEventHandler.h
```diff
@@ -0,0 +1,46 @@
...
+namespace PokemonAutomation{
+
+class ControllerState;  // forward declaration to avoid circular dependency
+
+class KeyboardEventHandler{
+public:
+    KeyboardEventHandler();
+    virtual ~KeyboardEventHandler();
+
+     struct KeyboardListener{
```
Mysticial, line 24: indent

SerialPrograms/Source/Controllers/KeyboardInput/KeyboardInput.h
```diff
@@ -122,11 +126,13 @@ class KeyboardManager : public KeyboardInputController{
         deltas.to_state(static_cast<StateType&>(state));
     }
     virtual void cancel_all_commands() override{
+        WallClock time_stamp = current_time();
         WriteSpinLock lg(m_lock);
         if (m_controller == nullptr){
             return;
         }
         m_controller->cancel_all_commands();
```
Mysticial, line 134: Move the timestamp read to just before this line since it's closer to the actual time being sent to the Switch. Having it before the lock could mean an arbitrary delay.

SerialPrograms/Source/NintendoSwitch/Controllers/NintendoSwitch_Joycon.cpp
```diff
@@ -78,6 +79,7 @@ class JoyconController::KeyboardManager final :
             switch_state.joystick_y,
```
Mysticial, line 79: Same thing here. Move the timestamp read to just before the command is sent.

Common/Cpp/Json/JsonObject.h
```diff
@@ -18,6 +18,13 @@ class JsonObject{
     JsonObject() = default;
     JsonObject(JsonObject&& x) = default;
     JsonObject& operator=(JsonObject&& x) = default;
+
+    bool operator==(const JsonObject& x){ // TODO: implement == properly. For JsonValue as well.
+        return dump() == x.dump();
+    }
+    bool operator!=(const JsonObject& x){
+        return dump() != x.dump();
```
Mysticial, line 26: These are more commonly written as a delegation: `return !(*this == *this);`

PR #775, opened 2025-10-28
Added non-template constructor

Ericzklm:
Compile issue on MacOS x86 runners. The DetectorToFinder constructors in PLZA files pass in more arguments and the constructor cannot be matched. Adding in a base constructor resolves the lookup to match the constructor to the correct template

### DIFF
--- a/SerialPrograms/Source/CommonTools/VisualDetector.h
+++ b/SerialPrograms/Source/CommonTools/VisualDetector.h
@@ -42,6 +42,17 @@ public:
         CONSISTENT, //  process_frame() returns true when detected consecutively or not detected consecutively
     };
 
+    DetectorToFinder(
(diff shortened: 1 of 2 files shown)


Mysticial:
What about this?
```
    using Parent = DetectorToFinder<ButtonDetector>;


    ButtonWatcher( 
        Color color, 
        ButtonType button_type, 
        const ImageFloatBox& box, 
        VideoOverlay* overlay = nullptr, 
        std::chrono::milliseconds hold_duration = std::chrono::milliseconds(250) 
    ) 
         : Parent("ButtonWatcher", hold_duration, color, button_type, box, overlay) 
    {} 
```

Mysticial:
I'll merge this, but this has got to be a bug in Clang.

PR #785, opened 2025-11-02
Add BerryBuyer; Shift Dialog Detection to avoid Japanese Furigana

NympheaR:
Adding BerryBuyer because sometimes it's faster than resetting everything at Rust Syndicate (so that HP EVs for alphas don't go to waste, etc.)
Shifted FlatWhiteDialogDetector top detection box due to Japanese text sometimes interfering with detection (see image)
2511_PR_Furigana_Box

### DIFF
--- a/SerialPrograms/Source/PokemonLZA/Inference/PokemonLZA_DialogDetector.cpp
+++ b/SerialPrograms/Source/PokemonLZA/Inference/PokemonLZA_DialogDetector.cpp
@@ -211,15 +211,15 @@ bool NormalDialogDetector::process_frame(const ImageViewRGB32& frame, WallClock 
 FlatWhiteDialogDetector::FlatWhiteDialogDetector(Color color, VideoOverlay* overlay)
     : m_color(color)
     , m_overlay(overlay)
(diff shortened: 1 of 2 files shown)


SerialPrograms/Source/PokemonLZA/Inference/PokemonLZA_DialogDetector.cpp
```diff
@@ -211,15 +211,15 @@ bool NormalDialogDetector::process_frame(const ImageViewRGB32& frame, WallClock
 FlatWhiteDialogDetector::FlatWhiteDialogDetector(Color color, VideoOverlay* overlay)
     : m_color(color)
     , m_overlay(overlay)
-    , m_top(0.267838, 0.784156, 0.467618, 0.019531)
+    , m_bottom(0.265, 0.930, 0.465, 0.020)
```
Mysticial: Btw, if you want to easily get box coordinates, you can use the Box Draw program. So if you're wondering why all the new programs have 6 digits of precision for all box coordinates it's because we're all being lazy and using Box Draw to create them for us.

PR #802, opened 2025-11-06
add program for shiny hunting static Alpha Whirlipede

naussika:


Mysticial:
I'm not sure if we should jump straight into a separate program for every single Pokemon since we're looking potentially a lot of them.

So we should discuss on how to handle this and how we could group similar programs into the same ones.

Mysticial:
WZ Entrance on Zone 5 hits this alpha Whirlipede. I got one today. lol

Mysticial:
Yeah, we should make a list of common hunts and the best way to hunt it. We'll continue on Discord.

PR #819, opened 2025-11-12
close_game_from_home: updated to handle the case where the game is not selected.

jw098:
Uses a newly created CloseGameDetector.

### DIFF
--- a/SerialPrograms/Source/NintendoSwitch/DevPrograms/TestProgramSwitch.cpp
+++ b/SerialPrograms/Source/NintendoSwitch/DevPrograms/TestProgramSwitch.cpp
@@ -290,10 +290,14 @@ void TestProgram::program(MultiSwitchProgramEnvironment& env, CancellableScope& 
     [[maybe_unused]] VideoFeed& feed = env.consoles[0];
     [[maybe_unused]] VideoOverlay& overlay = env.consoles[0];
     ProControllerContext context(scope, console.controller<ProController>());
+    // JoyconContext context(scope, console.controller<JoyconController>());
     VideoOverlaySet overlays(overlay);
 
 
-
+#if 0
+close_game_from_home(console, context);
+// ssf_issue_scroll(context, DPAD_DOWN, 24ms);
+#endif  
 
 #if 0
     // auto snapshot = feed.snapshot();


SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_GameEntry.cpp
```diff
@@ -116,10 +122,31 @@ void close_game_from_home(ConsoleHandle& console, ProControllerContext& context)
...
+            pbf_mash_button(context, BUTTON_A, 50);
+            ensure_at_home(console, context);
+        }else{
+            console.log("Game was already closed.");
+        }
+        
+    }
 
     // fail-safe against button drops and unexpected error messages.
     pbf_mash_button(context, BUTTON_X, 50);
```
jw098: Since we're now using visual inference, mashing X and B are no longer necessary, right? Should we remove them?

SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_GameEntry.cpp
```diff
@@ -116,10 +122,31 @@ void close_game_from_home(ConsoleHandle& console, ProControllerContext& context)
```
Mysticial: You're actually gonna want to split this function into to, a blind one with the old behavior and the new feedback one. Since some of the older SwSh programs are able to run without feedback.

So give this the same treatment as `start_game_from_home()`. Rename the current one to  `start_game_from_home_blind()`, then add the new feedback one.

SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_GameEntry.cpp
```diff
@@ -136,11 +163,31 @@ void close_game_from_home(ConsoleHandle& console, JoyconContext& context){
...
                                                         // if game initially open.  |  if game initially closed
     pbf_mash_button(context, BUTTON_X, 800ms);          // - Close game.            |  - does nothing
-    pbf_move_joystick(context, 128, 255, 100ms, 10ms);  // - Does nothing.          |  - moves selector away from the closed game to avoid opening it.
-    pbf_move_joystick(context, 128, 255, 100ms, 10ms);  // - Does nothing.          |  - Press Down a second time in case we drop one.
-    pbf_mash_button(context, BUTTON_A, 400ms);          // - Confirm close game.    |  - opens an app on the home screen (e.g. Online)
-    go_home(console, context);                          // - Does nothing.          |  - goes back to home screen.
+    context.wait_for_all_requests();
+    if (is_close_game_menu_visible(console)){  // game was initially open. 
+        console.log("Detected close game menu.");
+        pbf_mash_button(context, BUTTON_A, 400ms);
```
Mysticial: Is it possible to state machine this?

**close game menu:**
Press A

**main menu:**
if (close game seen) return;
if (main menu seen twice before) return;
if (main menu seen once before)
    move joysticks, press Home.
else
    mash X

SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_GameEntry.cpp
```diff
@@ -136,11 +163,31 @@ void close_game_from_home(ConsoleHandle& console, JoyconContext& context){
...
+            console.log("Detected close game menu.");
+            pbf_mash_button(context, BUTTON_A, 400ms);
+            ensure_at_home(console, context);
+        }else{
+            console.log("Game was already closed.");
+        }
 
+    }
     // fail-safe against button drops and unexpected error messages.
     pbf_mash_button(context, BUTTON_X, 400ms);
```
Mysticial: If we can state machine the above, we can get rid of these two and save 2+ seconds from the reset sequence.

SerialPrograms/Source/NintendoSwitch/Programs/NintendoSwitch_GameEntry.cpp
Mysticial: Oh yes. Commented below without seeing this one on top (on the otherwise identical code).

PR #839, opened 2025-11-20
Fix video rotation: compensate for 90deg counter-clockwise camera rot…

jimtempleton21:
…ation

- Rotate display 90deg clockwise in CameraWidgetQt6 and CameraWidgetQt6.5 backends to fix display orientation
- Rotate snapshots 90deg clockwise in SnapshotManager to fix aspect ratio validation
- Ensures display and inference snapshots match raw frame orientation


Mysticial:
Thanks for contributing! However, This hasn't been a problem for anyone else, and AFAICT, you're unconditionally rotating the display. So it's going to rotate it wrong for everyone else.

More comments inline in the code.

PR #853, opened 2025-11-24
Attempt to fix Jachinthe.

Mysticial:


Mysticial:
Nope, this doesn't work.

PR #859, opened 2025-11-26
ML: Add CPU fallback if GPU fails. LabelImages: allow user to select color of selected annotation.

jw098:


SerialPrograms/Source/ML/Inference/ML_YOLOv5Detector.cpp
```diff
@@ -89,7 +91,31 @@ bool YOLOv5Detector::detect(const ImageViewRGB32& screen){
     cv::cvtColor(frame_mat_bgra, frame_mat_rgb, cv::COLOR_BGRA2RGB);
     
     m_output_boxes.clear();
-    m_yolo_session->run(frame_mat_rgb, m_output_boxes);
+
+    // fall back to CPU if fails with GPU.
+    for(size_t i = 0;;i++){
```
Gin890: Same here. Don't use for infinite for loop when you can just have two calls of the session run, one for gpu, and one for cpu if exception is caught.

SerialPrograms/Source/ML/Models/ML_ONNXRuntimeHelpers.cpp
```diff
@@ -58,8 +58,7 @@ Ort::SessionOptions create_session_options(const std::string& model_cache_path){
     so.AppendExecutionProvider("CoreML", provider_options);
     std::cout << "Using CoreML execution provider for GPU acceleration" << std::endl;
 #elif _WIN32
-#define ENABLE_WIN_GPU
-#ifdef ENABLE_WIN_GPU
+if (use_gpu){
```
Gin890: APPLE's impelemention should be inside this if(use_gpu) as well

PR #892, opened 2025-12-09
TurboMacro: add "Action" column that summarizes a given row's action.

jw098:
The caveat is that now the Procontroller table is running out of width real estate in the UI. i.e. the table is getting too wide.
Ideally, the joystick columns would be narrower, so we have more space.

### DIFF
--- a/SerialPrograms/Source/NintendoSwitch/Controllers/Joycon/NintendoSwitch_JoyconTable.cpp
+++ b/SerialPrograms/Source/NintendoSwitch/Controllers/Joycon/NintendoSwitch_JoyconTable.cpp
@@ -53,10 +53,10 @@ void register_joycon_tables(){
         ControllerCommandTable::make_row<JoyconStateRow>,
         {
             "Milliseconds",
+            "Action",
             "Buttons",
             "Joystick (X)",
             "Joystick (Y)",
-            "Action",
         }
     );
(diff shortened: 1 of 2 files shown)


SerialPrograms/Source/NintendoSwitch/Controllers/Joycon/NintendoSwitch_JoyconTable.cpp
```diff
@@ -87,11 +95,19 @@ JoyconStateRow::JoyconStateRow(EditableTableOption& parent_table)
     )
     , JOYSTICK_X(LockMode::UNLOCK_WHILE_RUNNING, 128, 0, 255)
     , JOYSTICK_Y(LockMode::UNLOCK_WHILE_RUNNING, 128, 0, 255)
+    , ACTION(false, LockMode::UNLOCK_WHILE_RUNNING, "", "")
 {
     PA_ADD_OPTION(DURATION);
     PA_ADD_OPTION(BUTTONS);
     PA_ADD_OPTION(JOYSTICK_X);
     PA_ADD_OPTION(JOYSTICK_Y);
+    PA_ADD_OPTION(ACTION);
```
Mysticial: I would actually put this between `DURATION` and `BUTTONS`. IOW, put it near the beginning.

PR #950, opened 2025-12-26
Remove TICKS_PER_SECOND.

Mysticial:


jw098:
Actually, I removed TICKS_PER_SECOND in my other draft PR. So, I don't think we need this PR for now.

PR #989, opened 2026-01-01
fixes keyboard control for dpad: add mapping for additional number keys

naussika:


naussika:
updated the mapping to `IJKLUOM,`

PR #995, opened 2026-01-03
tweak Virizion hyperspace route

theastrogoth:
Detects the console type on program start, and slightly adjusts the timing of the Virizion route for better reliability on Switch 1.

These timings work consistently on my setup, and I left the Switch 2 timings from Gin intact. For jumping the gap between buildings, using ssf for mashing A leading up the jump might be a good idea, but I haven't done that here in case it causes an unintended change to Switch 2 behavior.


theastrogoth:
I've gone ahead and added the A-mashing for the jump to avoid any lingering timing issues that prevent successfully checking for the shiny

Gin890:
Thanks!

PR #1004, opened 2026-01-04
Donut Maker Modified Travelling

Ericzklm:
Changed fast travel logic for donut making to remove dependency on player location
1. The program opens the map, opens the fast travel menu with 'Y', and flies to whatever is currently selected in order to set a backup save
2. Do a reset from backup save to create the starting point
    a. This should also clear any map filter which would be a problem if you need to keep the first donut
3. Do donut iterations as normal except to get to Hotel Z, open the map with 'Y', open the menu with '-', and go 4 slots down
    a. Nothing is selected so the selector should always be at the first item (Centrico Plaza)
    b. The order should always be Centrico Plaza -> Gare de Lumiose -> Pokemon Research Lab -> Hotel Z, all of which are guaranteed to be available at this point
4. Fast traveling away from Hotel Z can go to Centrico Plaza since the routine to fly to Hotel Z is no longer coupled with Vert Pokecenter

This should allow the user to start the program anywhere without any setup, should not be affected by quest labels, and should make the program agnostic of the player's location

### DIFF
--- a/SerialPrograms/Source/PokemonLZA/Programs/Farming/PokemonLZA_DonutMaker.cpp
+++ b/SerialPrograms/Source/PokemonLZA/Programs/Farming/PokemonLZA_DonutMaker.cpp
@@ -375,6 +375,31 @@ void load_backup_save(SingleSwitchProgramEnvironment& env, ProControllerContext&
     }
 }
 
+void exit_menu_to_overworld(SingleSwitchProgramEnvironment& env, ProControllerContext& context){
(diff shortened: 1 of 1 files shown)


Gin890:
I'm still unclear on the reason for step 2: "Do a reset from backup save to create the starting point
a. This should also clear any map filter which would be a problem if you need to keep the first donut". Can you explain more on the reason for doing a reset from backup save before start making donuts?

PR #1006, opened 2026-01-07
Proposed pybind for the Pro Controller.

Mysticial:
When you wrap this in python:

`wait()` -> `pbf_wait()`
`push_button()` -> `ssf_press_button()`
`push_dpad_button()` -> `ssf_press_dpad()`
...

`pbf` wraps around `ssf` as usual.

### DIFF
--- a/SerialPrograms/Source/Integrations/PybindSwitchController.cpp
+++ b/SerialPrograms/Source/Integrations/PybindSwitchController.cpp
@@ -33,7 +33,7 @@ public:
     void wait_for_ready(){
         std::unique_lock<std::mutex> lg(m_wait_lock);
         m_cv.wait(lg, [this]{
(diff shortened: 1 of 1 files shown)


SerialPrograms/Source/Integrations/PybindSwitchController.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+
+
+
+class PybindSwitchProControllerInternal final : public ControllerConnection::StatusListener{
+public:
+    PybindSwitchProControllerInternal(const std::string& name)
+        : m_logger(global_logger_raw(), "[Pybind]")
```
Mysticial: `global_logger_raw()` has a dependency on the `FileWindowLogger` which is Qt. So you can probably replace this with some null logger or a logger that simply prints to stdout.

PR #1026, opened 2026-01-12
Fix Hyperspace Cobalion shiny check

theastrogoth:
The route for checking for shininess now runs forward a little bit further, based on a report from L Aj


Gin890:
Thanks for the fix!

PR #1031, opened 2026-01-13
Actually fix Cobalion shiny check

theastrogoth:
The addition of the `context.wait_for_all_requests()` had broken part of the route for checking Cobalion by changing the last part of the route from a sprint to a walk, which is probably what was responsible for some [reports ](https://discord.com/channels/695809740428673034/1435100073406103562/1460109084702543934) of [problems](https://discord.com/channels/695809740428673034/1435100073406103562/1460678550557425734) with the beta release.

The attempted fix in #1026 probably didn't help, but this one will basically return things to the route that I originally tested. 

I won't be able to retest shiny sound detection, but I can re-check the route once I have some time to roll the correct Hyperspace map.


Gin890:
Thank you for the fix!

PR #1036, opened 2026-01-14
PaddleOCR

jw098:


Gin890:
Thanks for adding PaddleOCR!
Some nitpick: can you change new function names to be "lower_case_with_underscore()"? This is the default function naming style for C++

SerialPrograms/Source/CommonTools/OCR/OCR_Routines.cpp
```diff
@@ -40,14 +44,27 @@ StringMatchResult multifiltered_OCR(
 
     double pixels_inv = 1. / (image.width() * image.height());
 
+    bool use_paddle_ocr = GlobalSettings::instance().USE_PADDLE_OCR;
+    std::unique_ptr<ML::PaddleOCRPipeline> paddle_ocr;
+    if (use_paddle_ocr) {
+        // Initialize only if the setting is enabled
+        paddle_ocr = std::make_unique<ML::PaddleOCRPipeline>(language);
```
Mysticial: How expensive is this constructor? Is it okay to reinitialize it on every OCR call? Or should we cache like Tesseract?

SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
```diff
@@ -241,6 +247,7 @@ GlobalSettings::GlobalSettings()
     PA_ADD_OPTION(STATS_FILE);
     PA_ADD_OPTION(TEMP_FOLDER);
     PA_ADD_OPTION(THEME);
+    PA_ADD_OPTION(USE_PADDLE_OCR);
```
Mysticial: Can you put this under dev mode?

Mysticial:
Can you rebase this? I'll merge, then we'll figure out the caching later.

SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.cpp
jw098: Visibility is already restricted to Dev mode, but you only see it under load_json. Can't do it in the constructor since the JSON hasn't been loaded yet.

PR #1038, opened 2026-01-15
Alternative PLZA Latias Program

ercdndrs:


SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_HyperspaceLegendary.cpp
```diff
@@ -82,6 +89,7 @@ ShinyHunt_HyperspaceLegendary::ShinyHunt_HyperspaceLegendary()
             {Legendary::COBALION, "cobalion", "Cobalion"},
             {Legendary::TERRAKION, "terrakion", "Terrakion"},
             {Legendary::VIRIZION,  "virizion",  "Virizion"},
+            {Legendary::LATIAS_ALT, "latias_alt", "Latias_Alt"},
```
Ericzklm: What about renaming the options to:
- Latias: Shuttle Run
- Latias: Game Reset
- Latios: Game Reset
- Cobalion: Shuttle Run
- Terrakion: Shuttle Run
- Virizion: Shuttle Run

The "min calories" option could then specify it's only needed for routes labeled Shuttle Run and it would present the routes as different flavors rather than main/alternative

PR #1057, opened 2026-01-19
File IO Class

Gin890:
Wraps FILE* for fast file IO!
Claude coded it so we need more tests on it

### DIFF
--- a/Common/Cpp/FileIO.cpp
+++ b/Common/Cpp/FileIO.cpp
@@ -1,0 +1,177 @@
+/*  File IO
+ *
+ *  From: https://github.com/PokemonAutomation/
+ *
+ */
+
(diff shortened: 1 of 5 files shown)


Common/Cpp/FileIO.cpp
```diff
@@ -0,0 +1,167 @@
...
+        return false;
+    }
+    return fflush(m_file) == 0;
+}
+
+long FileIO::tell() const{
+    if (!m_file){
+        return -1;
+    }
+    return ftell(m_file);
```
Mysticial: Ok this won't work on Windows since it's only 32-bit. So it will break on files larger than 4GB. (or maybe 2GB if it's signed?)

You need `_ftelli64()` for Windows.

PR #1063, opened 2026-01-20
Change the static global logger to be a UI-free logger

Gin890:
The static global logger is now a FileLogger. The old GUI-dependent logger, FileWindowLogger is removed. Each LogWindow is connected directly to the global FileLogger.

### DIFF
--- a/Common/Cpp/Logging/FileLogger.h
+++ b/Common/Cpp/Logging/FileLogger.h
@@ -29,10 +29,11 @@ struct FileLoggerConfig{
     size_t max_queue_size = 10000;                  // Max pending log entries before blocking
     size_t max_file_size_bytes = 50 * 1024 * 1024;  // Max file size before rotation (50MB default)
     size_t last_log_max_lines = 10000;              // Max lines to keep in memory for get_last()
+    bool flush_when_exit = true;                    // Whether to save remaining logs to file when program exists
 };
 
 
-// A Qt-free file logger that:
+// A file logger that:
 // 1. Writes log messages to a file asynchronously via a background thread
 // 2. Supports log rotation when the file exceeds a configured size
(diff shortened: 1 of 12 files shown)


Common/Cpp/Logging/FileLogger.h
```diff
@@ -28,10 +28,11 @@ struct FileLoggerConfig{
     size_t max_queue_size = 10000;                  // Max pending log entries before blocking
     size_t max_file_size_bytes = 50 * 1024 * 1024;  // Max file size before rotation (50MB default)
     size_t last_log_max_lines = 10000;              // Max lines to keep in memory for get_last()
+    bool flush_when_exit = false;                   // Whether to save remaining logs to file when program exists
```
Mysticial: Should default to true?

Common/Cpp/Logging/GlobalLogger.cpp
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#include "GlobalLogger.h"
+
+namespace PokemonAutomation{
+
+
+static FileLoggerConfig _global_logger_config;
+
+void initialize_global_logger(FileLoggerConfig config){
```
Mysticial: We need to be careful here that nothing calls the logger during static init.

Alternatively, you can make the entire logger lazy static init.

Gin890: How can you do that? The logger object is already lazy static init but it needs to be set with a proper logger config, which I can only implement in this PR by calling `initialize_global_logger()` at start of main() to initialize a static config struct before any logger calls.

Mysticial:
Pulled some of this out to: https://github.com/PokemonAutomation/Arduino-Source/commit/d00cf76a1216a5901baff861f99b802ca2b9b93a

PR #1079, opened 2026-02-06
add inference stop condition to LZA Turbomacro

jw098:


SerialPrograms/Source/PokemonLZA/Programs/PokemonLZA_TurboMacro.cpp
```diff
@@ -0,0 +1,122 @@
...
+//using std::endl;
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonLZA{
+
+
+LZA_TurboMacro_Descriptor::LZA_TurboMacro_Descriptor()
+    : SingleSwitchProgramDescriptor(
+        "NintendoSwitch:TurboMacro",
```
Mysticial: You're gonna want to update these to maybe, `ShinyHunt:Macro` or something like this. Then also add the usual stats and stop condition as the other shiny hunting programs.

PR #1085, opened 2026-02-09
Fix the loading time of Magearna stats reset is too short for NS1

ZHDreamer:
The loading time for NS1 is longer than NS2. The original program can't run on my NS1. So I add the overworld detection and A button detection to make the program more robust. Also add some duration for the dialogue.


Mysticial:
This one is pending some testing by a few people before we merge.

ZHDreamer:
> This one is pending some testing by a few people before we merge.

Sure, I've tested on my ns1 for 300 rounds without any problem. But it needs to be tested on ns2 to see if I break something.

PR #1087, opened 2026-02-11
Update Watt Farmer with non-NSO option

Boombaastical:
I find myself often in need of farming Watts without wanting to go through recreating another Nintendo Account, unlinking it, to get the free 7 days. And if you have multiple accounts you may just only use NSO on your main one. In any case, I'm directly using it now.


SerialPrograms/Source/PokemonSwSh/Programs/DateSpamFarmers/Original_PokemonSwSh_DateSpam-WattFarmer copy.h
Ericzklm: I believe this was accidentally added and should be removed

PR #1090, opened 2026-02-14
BDSP Money Farmer route 210: Movement according to which of the 2 trainers is available

Boombaastical:
Since I worked on that one to add pokemon with Pickup as an option in another PR, I noticed that the program often returns errors because it doesn't move correctly to the one trainer out of the two who is ready to battle. Fixed that, tried it with a few different trainer configurations and it works so far.

Not sure how the values I included would work with the Switch 2's resolution?


Mysticial:
Still doesn't work. At this point, I'm just going to close this PR since it hasn't made any progress in 4 months.

https://github.com/user-attachments/assets/16928a8c-3915-4934-9189-7bdc8b6f84df

PR #1098, opened 2026-02-25
File Unzip using miniz library

jw098:
Note: this does not extract tar files, only zip files.

### DIFF
--- a/SerialPrograms/Source/NintendoSwitch/DevPrograms/TestProgramComputer.cpp
+++ b/SerialPrograms/Source/NintendoSwitch/DevPrograms/TestProgramComputer.cpp
@@ -374,7 +374,7 @@ void TestProgramComputer::program(ProgramEnvironment& env, CancellableScope& sco
 
 
 
-#if 1
+#if 0
     std::string zip_path = SETTINGS_PATH() + "test2.zip";
     std::string target_dir = SETTINGS_PATH() + "testB/";
     unzip_file(zip_path.c_str(), target_dir.c_str());


Mysticial:
Do we know how well this handles really large files?

Say you try to unzip a 10GB file. Does it store the whole thing in memory? (10GB memory usage)

Are there any progress indicators while it's decompressing a large file? Or does it just hang until it returns?

jw098:
I tested it on a 16GB file and it does work. It prints out progress as it unzips. It's not currently done on a background thread. But I feel that can be added later.

SerialPrograms/Source/CommonFramework/Tools/FileUnzip.cpp
```diff
@@ -0,0 +1,203 @@
...
+    bool is_safe(const std::string& target_dir, const std::string& entry_name) {
+        try {
+            // 1. Get absolute, normalized paths
+            // handles symlinks. and resolves .. and . components. throws error if path doesn't exist
+            Filesystem::Path base = fs::canonical(Filesystem::Path(target_dir));
+            // confirms that base is a directory, and not a file
+            if (!fs::is_directory(base)) return false;
+            
+            // resolves .. and . components and returns an absolute path without requiring the final path to exist.
+            fs::path target = fs::weakly_canonical(Filesystem::Path(base / entry_name));
```
jw098: Technically, this is also a usage of std::filesystem::path (same with the use of canonical a few lines above). Although it isn't constructing a path from a string. Do need to move this to PokemonAutomation::Filesystem::Path as well?

PR #1101, opened 2026-02-27
Max lair/added save on the go

Boombaastical:
As I often ran these Max Lair adventures, I thought that stopping on each new boss that one wants to save takes quite a bit more time than saving them automatically on the go. This would be especially interesting for those who just start with the Max Lair and keep the paths for the difficult bosses (e.g. Zygarde) directly.

Not sure if the code is proper, I tried to use the same style as was there before with a while loop. I tested it a few times and it works smoothly, I added extra time to make sure not to lose button presses because it's quite an important task.


SerialPrograms/Source/PokemonSwSh/MaxLair/Framework/PokemonSwSh_MaxLair_StateMachine.cpp
```diff
@@ -182,16 +182,18 @@ StateMachineAction run_state_iteration(
...
-            console, context,
-            save_path,
-            global_state
-        );
+                     runtime,
+                     env, console_index,
+                     console, context,
+                     save_path,
+                     global_state
+                     );
```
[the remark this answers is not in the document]
Boombaastical: Updated this, I thought I reverted back to the original, should've checked before sending the PR, sorry.

SerialPrograms/Source/PokemonSwSh/MaxLair/Options/PokemonSwSh_MaxLair_Options_BossAction.cpp
```diff
@@ -15,6 +15,10 @@
 #include "PokemonSwSh/Resources/PokemonSwSh_PokemonSprites.h"
 #include "PokemonSwSh/Resources/PokemonSwSh_MaxLairDatabase.h"
 #include "PokemonSwSh_MaxLair_Options_BossAction.h"
+#include "Common/Cpp/Options/BooleanCheckBoxOption.h"
+#include "Common/Cpp/Options/ConfigOption.h"
+#include <vector>
+#include <memory>
```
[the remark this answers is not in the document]
Boombaastical: Now should be the right logic, indeed I didn't notice the order, but thanks for clarifying!

SerialPrograms/Source/PokemonSwSh/MaxLair/Program/PokemonSwSh_MaxLair_Run_Entrance.cpp
```diff
@@ -8,23 +8,82 @@
...
+                                          Language language,
+                                          const ImageViewRGB32& screen
+                                          ) {
+    
+    std::vector<std::string> slugs;
+    
+    for (int i = 0; i < 3; ++i) {
+        auto cropped = extract_box_reference(screen, NAME_BOXES[i]);
+        OCR::StringMatchResult result = PokemonNameReader::instance().read_substring(
+                                                                                     stream.logger(), language, cropped, OCR::BLACK_OR_WHITE_TEXT_FILTERS(),
```
Mysticial: formatting

Boombaastical:
Updated with some new changes, still draft

Boombaastical:
Updated the PR so as to change only so that the user can select any number of paths it wants to save, and then stops when the list is full. Also, corrected some formatting and improved the notifications sent (now only sends when encountering a new boss, not when saving a path that was already saved before).

SerialPrograms/Source/PokemonSwSh/MaxLair/Program/PokemonSwSh_MaxLair_Run_Entrance.cpp
```diff
@@ -36,23 +39,101 @@ void run_entrance(
...
-            break;
+        if (!screen) continue;
+        
+        ImageStats dialog_box_stats = image_stats(extract_box_reference(screen, dialog_box));
+        ImageStats yes_no_box_stats = image_stats(extract_box_reference(screen, yes_no_box));
+        ImageStats paths_box_stats = image_stats(extract_box_reference(screen, paths_box));
+        
+        bool dialog_box_present = is_grey(dialog_box_stats, 400, 1000);
+        bool yes_no_box_present = is_white(yes_no_box_stats, 400, 10);
+        bool paths_box_present = is_white(paths_box_stats, 400, 10);
```
Mysticial: I started looking at this again since my other save is ready to do DAs.

Please use the proper dialog detectors instead of doing your own thing.

- `PokemonSwSh::BlackDialogBoxDetector`
- `PokemonSwSh::SelectionArrowFinder`

Boombaastical:
Updated with the selectionArrowFinder. It now detects 2 areas, if that selectionArrowFinder is in the top area + a small white strip, it means that the paths are full, if it detects it in the bottom area, it's the yes/no prompt, if none of these are true, it continues pressing A until the dialogue disappears (as is the case now).
Screenshot 2026-05-05 at 19 20 49

SerialPrograms/Source/PokemonSwSh/MaxLair/Program/PokemonSwSh_MaxLair_Run_Entrance.cpp
```diff
@@ -36,23 +41,109 @@ void run_entrance(
...
+        if (!boss_slug.empty()){
+            save_path = runtime.actions.is_in_save_list(boss_slug);
+            stream.log("Boss: " + boss_slug + ", should save: " + (save_path ? "Yes" : "No"), COLOR_BLUE);
+        }
+    }else{
+        save_path = followed_path;
+    }
+    
+    // Selection arrow in the path-list area — present when the "list is full / replace?" dialog is active
+    SelectionArrowFinder top_region(stream.overlay(), {0.63, 0.50, 0.22, 0.10});
```
Mysticial: Reviewing this today. One concern I have about this line is that the length varies depending on the language.

I'm actually working on a different way to write this loop that's more consistent with the newer programs. (you might have noticed that I've pushed some SwSh detector stuff to main) So I'll probably push a few things to this branch later.

PR #1110, opened 2026-03-02
Max Lair - Added the option to stop on Alolan Raichu and Alolan Marowak

Boombaastical:
As notified two days ago over Discord, the idea is simply to be able to stop on certain pokemon encountered during the Run_Battle part, so that the change doesn't become so invasive (I also don't want to start an overhaul of the whole MaxLair program with my limited C++ knowledge).

I also left an option to include the whole list of pokemon instead of Alolan Raichu and Marowak, but since the list is rather long, I don't think it makes much sense to add the whole list by default (for developers or those who would like this function specifically, they could still change the code by changing LIMITED_MODE in Options_NonBossAction.cpp.

I tested it before with the whole list to stop on certain pokemon, and it indeed stops as needed.


Mysticial:
Closing due to high risk and little benefit.

PR #1117, opened 2026-03-06
Refactor FRLG to use game cropping.

Mysticial:


kichithewolf (approved):
Tested swapping devices (black border check disabled since no home screen on the handheld): https://youtu.be/mwt2DWyiFF8

PR #1129, opened 2026-03-10
FRLG add selection arrow detection and start menu navigation

Dalton-V:
Selection arrow detection and saving tested on Switch1 and Switch2 with different capture cards.

### DIFF
--- a/SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_SelectionArrowDetector.cpp
+++ b/SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_SelectionArrowDetector.cpp
@@ -13,25 +13,33 @@ namespace PokemonAutomation{
 namespace NintendoSwitch{
 namespace PokemonFRLG{
 
-ImageFloatBox SelectionArrowDetector::arrow_box_for_position(SelectionArrowPosition position){
+ImageFloatBox SelectionArrowDetector::arrow_box_for_position(SelectionArrowPositionStartMenu position){
     switch (position){
-    case SelectionArrowPosition::START_MENU_POKEDEX:
+    case SelectionArrowPositionStartMenu::POKEDEX:
         return ImageFloatBox(0.685, 0.055, 0.03, 0.075);
-    case SelectionArrowPosition::START_MENU_POKEMON:
+    case SelectionArrowPositionStartMenu::POKEMON:
         return ImageFloatBox(0.685, 0.145, 0.03, 0.075);
-    case SelectionArrowPosition::START_MENU_BAG:
+    case SelectionArrowPositionStartMenu::BAG:
         return ImageFloatBox(0.685, 0.235, 0.03, 0.075);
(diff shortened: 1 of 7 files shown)


SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_SelectionArrowDetector.cpp
```diff
@@ -0,0 +1,140 @@
...
+    : m_color(color)
+    , m_overlay(overlay)
+    , m_arrow_box(arrow_box_for_position(position))
+{
+}
+void SelectionArrowDetector::make_overlays(VideoOverlaySet& items) const{
+    items.add(m_color, m_arrow_box);
+}
+bool SelectionArrowDetector::detect(const ImageViewRGB32& screen){
+    double screen_rel_size = (screen.height() / 1080.0);
```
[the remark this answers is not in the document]
Dalton-V: I think I followed for the most part. Updated and ran a quick test that the save function still works.

I don't think I follow the need to change the 3rd and 4th parameters for the image box as these are height and width. If we are just moving the boxes if the game box moves, shouldn't the size of these boxes stay the same?

SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_SelectionArrowDetector.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    double screen_rel_size = (game_screen.height() / 1080.0);
+    double screen_rel_size_2 = screen_rel_size * screen_rel_size;
+
+    double min_area_1080p = 700;
+    double rmsd_threshold = 80;
+    size_t min_area = size_t(screen_rel_size_2 * min_area_1080p);
+
+    const std::vector<std::pair<uint32_t, uint32_t>> FILTERS = {
+        {0xff464646, 0xff787878}
```
Mysticial: I bet we'll need more filters later. But we'll add them when we need to.

SerialPrograms/Source/PokemonFRLG/Programs/PokemonFRLG_StartMenuNavigation.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    return true;
+}
+
+void save_game_to_overworld(ConsoleHandle& console, ProControllerContext& context){
+
+    bool seen_start_menu = false;
+
+    WallClock start = current_time();
+    while (true){
```
Mysticial: Rather than using a bunch of seen flags, I'd do it this way:
```
while (true){
    if (too many attempts){
        // throw exception
    }

    //  mash B to back out to overworld.

    //  Press X and wait for start menu to show up
    if (start menu fails to show up){
        print error
        continue;
    }

    //  press + and wait for save dialog
    if (save dialog doesn't show up){
        print error
        continue;
    }

    //  repeat for every step...
}
```
Basically the moment anything unexpected happens, back out and try again. Don't bother with success flags that you need to jump over and such.

PR #1149, opened 2026-03-29
Add FGLR Lucky Egg Farmer

Dalton-V:
This adds a Lucky Egg Farmer for FRLG

### New Detectors
- `PartyHeldItemDetector` used to detect if a captured Chansey is holding a Lucky Egg
- `BattleSelectionArrowDetector` the selection arrow in battles is slightly different from the start menu arrow. This new arrow is also used when asked for a nickname.
- `PokedexRegisteredDetector` this detects the "A" button in the bottom right when a new pokemon is added to the dex. (Used if the player has not captured Chansey before.

### Updates to existing logic
- Safari Zone has a different start menu layout from the standard overworld.
- Updated the Start Menu navigation to search the whole area for the selection arrow instead of multiple watchers as mentioned in a previous PR.

### DIFF
--- a/SerialPrograms/Source/PokemonFRLG/Programs/Farming/PokemonFRLG_LuckyEggFarmer.cpp
+++ b/SerialPrograms/Source/PokemonFRLG/Programs/Farming/PokemonFRLG_LuckyEggFarmer.cpp
@@ -299,6 +299,8 @@ bool LuckyEggFarmer::attempt_catch(SingleSwitchProgramEnvironment& env, ProContr
 
         PokedexRegisteredWatcher pokedex_registered(COLOR_RED, &env.console.overlay());
 
+        WhiteDialogWatcher in_safari_zone_building(COLOR_RED);
+
         WallClock start = current_time();
         while (true)
         {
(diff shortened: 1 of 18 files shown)


Mysticial:
rebase

SerialPrograms/Source/PokemonFRLG/Programs/Farming/PokemonFRLG_LuckyEggFarmer.cpp
```diff
@@ -0,0 +1,517 @@
...
+        {
+            Language::English,
+            Language::Japanese,
+            Language::Spanish,
+            Language::French,
+            Language::German,
+            Language::Italian,
+            Language::Korean,
+            Language::ChineseSimplified,
+            Language::ChineseTraditional,
```
Mysticial: Same as the other PR. Are all these languages available?

Mysticial:
I haven't reviewed the actual Lucky Egg file yet. But all the detectors are fine if you want to pull those out into a separate PR.

SerialPrograms/Source/PokemonFRLG/Programs/Farming/PokemonFRLG_LuckyEggFarmer.cpp
Dalton-V: Updated to match EV program

PR #1155, opened 2026-04-05
Pokopia Cloud Island Reset

Ericzklm:
If Pokopia automation is of interest, this adds the start of the stack and a program to do (Mew) Stamp hunting using Cloud Island resets. 

The Cloud Island Reset is fairly destructive, it repeatedly deletes the Cloud Island which many people might actually be using, so it's locked behind dev mode.

This should provide the groundwork for some possible follow up ideas:
- Auto reset and purchase recipes via Cloud Island
- Local coin farmer by time traveling
- Local recipe farmer by time traveling

Needs cleanup and further testing, but open for review


Mysticial:
I'm in favor of adding Pokopia programs. Feel free to move this out of draft when you're ready.

Mysticial:
Since I don't play this game, I'm rubber stamping. You're in charge of it. :)

PR #1165, opened 2026-04-08
Resource Download table

jw098:
This only adds the ability to download the resource from the Settings page. It doesn't yet prompt you when you start a program. 

Another note: right now the URL to the resources points to my own repo. I assume we will want the URLs pointing to the official Packages releases.


SerialPrograms/Source/CommonFramework/Tools/FileUnzip.h
```diff
@@ -8,10 +8,20 @@
...
-void unzip_file(const char* zip_path, const char* target_dir);
+// unzips the zip file located in zip_path, to target_dir
+// if target_dir doesn't already exist, it will create it.
+// throw OperationCancelledException if the is_cancelled callback returns true
+// throw InternalProgramError if unzipping fails.
+void unzip_file(
+    const char* zip_path, 
+    const char* target_dir, 
+    std::function<void(int)> progress_callback,
+    std::function<bool()> is_cancelled
```
Mysticial: Without having looked much at the surrounding code, how difficult do you think it would be to re-architect this to use the `Cancellable(Scope)` class? Since that was designed specifically for this kind of async cancel stuff.

SerialPrograms/Source/CommonFramework/VideoPipeline/VideoOverlay.h
```diff
@@ -94,7 +94,7 @@ class VideoOverlay{
     void issue_mouse_move(double x, double y);
 
     struct KeyEventListener{
-        virtual void on_key_press(QKeyEvent* event){}
+        virtual void on_key_press(QKeyEvent* event){};
```
Mysticial: You don't need these semi-colons. Not sure how they got there in the first place.

SerialPrograms/Source/CommonFramework/Tools/FileHash.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <string>
+#include <functional>
+#include "Common/Cpp/CancellableScope.h"
+
+
+namespace PokemonAutomation{
+
+// uses SHA 256
+std::string hash_file(CancellableScope& scope, const std::string& file_path, std::function<void(int)> hash_progress);
```
Mysticial: I'd prefer the `hash_process` be a floating-point value between 0.0 - 1.0. Then have the display choose how to format it.

SerialPrograms/Source/CommonFramework/Tools/FileHash.h
```diff
@@ -0,0 +1,21 @@
+/*  File Hash
+ *
+ *  From: https://github.com/PokemonAutomation/
+ *
+ */
+
+#ifndef PokemonAutomation_FileHash_H
```
Mysticial, line 7: Since this PR is so big, let's chip away at it little at a time. This file can go in by itself.

SerialPrograms/Source/CommonFramework/Tools/FileHash.cpp
```diff
@@ -0,0 +1,61 @@
+/*  File Hash
+ *
+ *  From: https://github.com/PokemonAutomation/
+ *
+ */
+
+
```
Mysticial: A bit too many spaces here?

SerialPrograms/Source/CommonFramework/Tools/FileDownloader.h
```diff
@@ -17,6 +21,16 @@ namespace FileDownloader{
 //  Throws OperationFailedException if failed to download.
 std::string download_file(Logger& logger, const std::string& url);
 
+//  Throws OperationFailedException if failed to download.
+void download_file_to_disk(
+    CancellableScope& scope,
+    Logger& logger, 
+    const std::string& url, 
+    const std::string& file_path, 
+    qint64 expected_size,
```
Mysticial: Use `uint64_t` do you don't pull in Qt in this header. We'll hide all the Qt in the .cpp since those can be theoretically replaced with something else in the future.

SerialPrograms/Source/CommonFramework/Tools/FileHash.h
Mysticial: Actually no. Instead of a float. Let's pass both the # of bytes completed and # of total bytes. Since some fancy displays will want to show the breakdown.

SerialPrograms/Source/CommonFramework/Tools/FileDownloader.h
```diff
@@ -17,6 +21,16 @@ namespace FileDownloader{
...
 std::string download_file(Logger& logger, const std::string& url);
 
+//  Throws OperationFailedException if failed to download.
+void download_file_to_disk(
+    CancellableScope& scope,
+    Logger& logger, 
+    const std::string& url, 
+    const std::string& file_path, 
+    qint64 expected_size,
+    std::function<void(int)> progress_callback
```
Mysticial: Same as the other place. Pass both # of bytes completed and # of bytes total.

SerialPrograms/Source/CommonFramework/Options/LabelCellOption.h
```diff
@@ -45,6 +45,8 @@ class LabelCellOption : public ConfigOptionImpl<LabelCellOption>{
     const ImageViewRGB32& icon() const;
     Resolution resolution() const;
 
+    void set_text(std::string x);
```
Mysticial, line 48: The `LabelCellOption` and `LabelCellWidget` can also be chipped out and merged separately.

SerialPrograms/Source/CommonFramework/Options/QtWidget/LabelCellWidget.cpp
Mysticial, line 22: Detach listener here or it will crash if the option does an update after the UI has been closed.

`m_value.remove_listener(*this);`

SerialPrograms/Source/CommonFramework/Tools/FileHash.h
jw098: Right now, I'm limiting the rate of `hash_progress` updates, so it only fires when it reaches the next percentage.

I can update `hash_file` so that it will fire hash_progress very frequently (every 1MB) and the calling function can control how often to update its display. Is this what you'd prefer?

EDIT: answered on discord

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadHelpers.cpp
```diff
@@ -0,0 +1,141 @@
...
+    std::vector<DownloadedResourceMetadata> resources;    
+
+    try{
+        const JsonObject& obj = json.to_object_throw();
+        const JsonArray& resource_list = obj.get_array_throw("resourceList");
+        for (const JsonValue& resource_val : resource_list){
+            const JsonObject& resource_obj = resource_val.to_object_throw();
+
+            std::string resource_name = resource_obj.get_string_throw("resourceName");
+            std::optional<uint16_t> version_num = (uint16_t)resource_obj.get_integer_throw("version");
```
Mysticial: Can we change the naming convention to TitleCase?

- `ResourceList`
- `ResourceName`
- `Version`

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadHelpers.h
```diff
@@ -0,0 +1,61 @@
+/*  Resource Download Helpers
```
Mysticial, line 1: These two files `ResourceDownloadHelpers.h/cpp` are good to merge separately.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadOptions.h
```diff
@@ -0,0 +1,89 @@
...
+// #include "ResourceDownloadHelpers.h"
+
+
+namespace PokemonAutomation{
+
+class ResourceDownloadRow;
+
+
+class ResourceDownloadButton : public QObject, public ConfigOptionImpl<ResourceDownloadButton>{
+    Q_OBJECT
```
Mysticial: I don't see anything in this class that needs Qt. So I think you can remove this and the header.

This is an Option class which shouldn't have any Qt anyway.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.h
```diff
@@ -0,0 +1,147 @@
...
+#include "ResourceDownloadHelpers.h"
+#include "ResourceDownloadOptions.h"
+#include <mutex>
+
+namespace PokemonAutomation{
+
+
+// must be initialized as shared_ptr, so that `shared_from_this` can work
+// so you're forced to use the factory method `create`
+class DownloadThread : public CancellableScope, public std::enable_shared_from_this<DownloadThread> {
```
Mysticial: This is the first time I've heard of `std::enable_shared_from_this`. So I need to think about this.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.h
```diff
@@ -0,0 +1,147 @@
...
+        bool is_downloaded,
+        std::optional<uint16_t> version_num,
+        ResourceVersionStatus version_status
+    );
+
+public:
+    struct DownloadListener{
+        virtual void on_download_progress(uint64_t bytes_done, uint64_t total_bytes){};
+        virtual void on_unzip_progress(uint64_t bytes_done, uint64_t total_bytes){};
+        virtual void on_hash_progress(uint64_t bytes_done, uint64_t total_bytes){};
```
Mysticial: No semicolons needed here since these include definitions.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.cpp
```diff
@@ -0,0 +1,569 @@
...
+void ResourceDownloadRow::update_table_label(bool success){
+    set_is_downloaded(success);
+    set_version_status(success ? ResourceVersionStatus::CURRENT : ResourceVersionStatus::NOT_APPLICABLE);
+}
+
+
+
+ResourceDownloadRow::~ResourceDownloadRow(){
+    // cout << "~ResourceDownloadRow" << endl;
+    m_worker1.wait_and_ignore_exceptions();
```
Mysticial: What happens if the user closes the program which causes this to get destructed? If there are pending downloads/deletes, it waits for them to finish right?

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.h
jw098: My goal was to have the DownloadThread clean itself up once it finishes. When DownloadThread finishes its action, I wanted DownloadRow to release ownership, and for DownloadThread to delete itself.
If there's a cleaner way to do it, I'm open to suggestions.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.cpp
jw098: From my testing, I'm actually able to close the program mid-download without issues.

How do we shut down the thread, if we want it to stop right away?

Mysticial: You'd send a cancellation into it at the start of the destructor. I'll take another look at this tomorrow.

Mysticial: Yes, from what I can tell, you'd cancel `m_download_thread` here. That way it doesn't forcibly wait for the download to finish.

The cleanup thread probably won't take very long unless it gets stuck on an I/O error.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.h
```diff
@@ -0,0 +1,147 @@
...
+    struct Data;
+    Pimpl<Data> m_data;
+
+    ResourceDownloadButton m_download_button;
+    ResourceDeleteButton m_delete_button;
+    ResourceCancelButton m_cancel_button;
+    ResourceProgressBar m_progress_bar;
+
+    AsyncTask m_worker1;
+    AsyncTask m_worker2;
```
Mysticial: Probably want to rename `m_worker1` to `m_download_thread` and `m_worker2` to `m_delete_thread`?

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.cpp
```diff
@@ -0,0 +1,569 @@
...
+// DownloadThread
+/////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+DownloadThread::~DownloadThread(){
+    // cout << "~DownloadThread" << endl;
+    this->cancel();
+    m_worker.wait_and_ignore_exceptions();
+}
+DownloadThread::DownloadThread(ConstructorKey, ResourceDownloadRow& row) 
+    : CancellableScope()
```
Mysticial: Can omit this since you're not passing any parameters to it.

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownloadRow.h
Mysticial: Now that I've taken a look at this. I don't think you need to do this. You don't need to dangle the thread and have it clean itself up. You can just let the thread finish. And the `ResourceDownloadRow` can hold onto the thread handle until destruction. There's no harm for leaving the handle alive. And is standard practice anyway.

But there's a separate thing I want to bring up. What happens if the user initiates multiple downloads. Say they click download. But before the current download has finished, they click another one?

IOW, `start_download_thread()` is called while the thread is already active. From what I can tell, it will block until the existing thread finishes? Then you have a race condition with `on_download_finished()` to set/reset the pointer?

The way I'd architect this is to leave the download thread alive until `ResourceDownloadRow` is destructed. You create a data structure for things that need to be downloaded. When the user starts a download, you add the thing to the data structure and signal the thread to wake up and begin processing it. Once the data structure is empty, the thread sleeps. On destruction, you wake up the thread so it can exit naturally and join it.

I'm not sure how much of this design pattern you're familiar with. But I can walk you through this on Discord if needed.

jw098:
Closing this.
Replaced by PR #1272.

PR #1173, opened 2026-04-13
FRLG SID program

theastrogoth:
This adds a program to help users set their SID.

The SID is determined by the number of RNG advances between finalizing the player character's name and finishing the Professor Oak dialogue. The user manually enters their player name, and then starts the program before confirming it. 

SID can't be directly checked without custom firmware or ACE (which isn't available until pretty far into the game). Getting a shiny starter is typically how SID is confirmed, so testing for this will probably go hand in hand with an autonomous shiny starter manip program.

Depends on #1171 and #1172


theastrogoth:
This should probably go out at the same time as the StarterRNG program, since most users will need a way to determine their SID. 

This program in particular would benefit from testing by someone who can use CFW to directly see what SID they've hit. Preliminary testing on my end points to a bigger miss than I expected.

PR #1177, opened 2026-04-15
Add day/night detection and weather detection to za benchsit

Developer-Butters:
This PR adds a new Day/Night detector for Pokémon Legends: Z-A and integrates it into the Bench Sit shiny hunting program. I also added weather detection from the weather finder.

The detector samples color data directly from the main map to determine if the cycle is day or night. 

The Bench Sit program now allows the user to define specific weather and time conditions before the character runs forward. this should allow the user to more effectively shiny hunt specific Pokemon like Lucario.


SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -32,13 +35,13 @@ using namespace Pokemon;
 ShinyHunt_BenchSit_Descriptor::ShinyHunt_BenchSit_Descriptor()
     : SingleSwitchProgramDescriptor(
         "PokemonLZA:ShinyHunt-BenchSit",
-        STRING_POKEMON + " LZA", "Bench Sit",
-        "Programs/PokemonLZA/ShinyHunt-BenchSit.html",
-        "Shiny hunt by repeatedly sitting on a bench to reset spawns.",
```
Mysticial, line 37: Can you disable your code formatter? It really likes to mess everything up.

Developer-Butters: yes, I'll go grab the original file this evening and re-add my code once I disable. I tried to fix it manually last night but it didn't work and it was too late for me to keep changing stuff

Developer-Butters: fixed the formatting with the most recent push, but I need to do more testing to confirm some errors relating to the detector. for some reason the detector appears to be moving off the screen and causing a memory issue. Marking this PR as a draft for the time being.

Developer-Butters:
I ran the program for about 8 hours and didn't get the QT creator crash I had been getting before, so I believe this is ready for review.

Mysticial:
This doesn't seem to work. I tested this with day time only and it ran on both day and night.

https://github.com/user-attachments/assets/2ea1b479-5646-4bd4-b8c5-a052fd3fb084

Developer-Butters:
> This doesn't seem to work. I tested this with day time only and it ran on both day and night.
> 
>  2026-04-18.08-51-11.mov

I've made some changes in this most recent commit set. I'm letting it run for a bit to see if it works for me.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -92,6 +97,38 @@ ShinyHunt_BenchSit::ShinyHunt_BenchSit()
         100,
         0
     )
+    , DAY_NIGHT_FILTER(
+          "Run Forward Only During Day/Night:",
+          LockMode::LOCK_WHILE_RUNNING,
+          GroupOption::EnableMode::DEFAULT_DISABLED
+          )
+    , FILTER_MODE(
```
Mysticial: Rename this to `DAY_FILTER_MODE`.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -92,6 +97,38 @@ ShinyHunt_BenchSit::ShinyHunt_BenchSit()
...
+          LockMode::LOCK_WHILE_RUNNING,
+          GroupOption::EnableMode::DEFAULT_DISABLED
+          )
+    , FILTER_MODE(
+          "Time filter",
+          {
+           {0, "day", "Day"},
+           {1, "night", "Night"},
+           },
+          LockMode::LOCK_WHILE_RUNNING,
```
Mysticial: Since this is being accessed atomically, you can unlock it.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -92,6 +97,38 @@ ShinyHunt_BenchSit::ShinyHunt_BenchSit()
...
+          "Weather filter",
+          {
+           {0, "clear", "Clear"},
+           {1, "sunny", "Sunny"},
+           {2, "rain", "Rain"},
+           {3, "cloudy", "Cloudy"},
+           {4, "foggy", "Foggy"},
+           {5, "rainbow", "Rainbow"},
+           },
+          LockMode::LOCK_WHILE_RUNNING,
```
Mysticial: This can also be unlocked.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -92,6 +97,38 @@ ShinyHunt_BenchSit::ShinyHunt_BenchSit()
...
+          0
+          )
+    , WEATHER_FILTER(
+          "Run Forward Only During Specific Weather:",
+          LockMode::LOCK_WHILE_RUNNING,
+          GroupOption::EnableMode::DEFAULT_DISABLED
+          )
+    , WEATHER_FILTER_MODE(
+          "Weather filter",
+          {
```
Mysticial: Formatting here is inconsistent with everything else. Indentation is always 4 spaces. Closing braces always indent back 4 to align with where it opened if it opened on a new line.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -156,7 +197,61 @@ void run_back_until_found_bench(
...
+    
+    if (!m_day_night_detector){
+        m_day_night_detector = std::make_unique<DayNightStateDetector>(&overlay);
+    }
+    
+    if (!m_day_night_detector->detect(frame)){
+        return true;
+    }
+    DayNightState current_state = m_day_night_detector->state();
+  if (current_state == DayNightState::NIGHT) {
```
Mysticial: formatting

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -199,30 +294,132 @@ void ShinyHunt_BenchSit::program(SingleSwitchProgramEnvironment& env, ProControl
...
+                    run_back_until_found_bench(env, context);
+                    shiny_sound_handler.process_pending(context);
+                    continue;
+                }
+
+
+                // movement
                 if (duration > Milliseconds::zero()){
-                    if (WALK_DIRECTION.current_value() == 0){ // forward
+
```
Mysticial: Why are you removing comments?

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -199,30 +294,132 @@ void ShinyHunt_BenchSit::program(SingleSwitchProgramEnvironment& env, ProControl
...
                         env.console.overlay().add_log("Move Forward");
+
                         ssf_press_button(context, BUTTON_B, 0ms, 2*duration, 0ms);
+
                         pbf_move_left_joystick(context, {0, +1}, duration, 0ms);
-                        // run back
-                        pbf_move_left_joystick(context, {0, -1}, duration + 750ms, 0ms);
+
+                        pbf_move_left_joystick(context, {0, -1}, duration + 1000ms, 0ms);
```
Mysticial: Why are you adding newlines everywhere when it's only a 1 line change?

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -199,30 +294,132 @@ void ShinyHunt_BenchSit::program(SingleSwitchProgramEnvironment& env, ProControl
...
+
                         run_back_until_found_bench(env, context);
-                    }else if (WALK_DIRECTION.current_value() == 1){ // left
+                    }
+
```
Mysticial: formatting

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
```diff
@@ -199,30 +294,132 @@ void ShinyHunt_BenchSit::program(SingleSwitchProgramEnvironment& env, ProControl
...
                         pbf_move_left_joystick(context, {+1, 0}, duration, 0ms);
-                        pbf_press_button(context, BUTTON_L, 100ms, 400ms);
+
+                        pbf_press_button(context, BUTTON_L, 130ms, 450ms);
+
                         ssf_press_button(context, BUTTON_B, 0ms, duration, 0ms);
-                        pbf_move_left_joystick(context, {0, -1}, duration, 0ms);
-                        pbf_move_left_joystick(context, {+1, 0}, 100ms, 0ms);
+
+                        pbf_move_left_joystick(context, {0, -1}, duration + 150ms, 0ms);
```
Mysticial: Can you explain what adding 150ms does here?

Mysticial:
Now at least the day/night detection works. (I haven't checked the weather yet.)

But my main concern now is how long it takes to open/check the map. You're adding 6 seconds to open the map. This is done on every day night cycle. So for every day/night pair, you add 12 seconds of overhead. This means that any run duration less than 6 seconds will actually be slower by filtering. The majority of the use cases for walking forward are less than that.

IOW, the day/night detection needs to be a lot faster for this to be useful. Weather is a different story though. I haven't run the numbers on that to see whether it makes sense.

If the map detection can't be sped up, you'll need do something else. For day/night, it's easy since they always alternate between day/night. Do the detection at the beginning and remember what the parity it. Then only recheck maybe every 100 bench sits to make sure you didn't miss one and end up on the wrong parity.

For weather detection, that one is tricker. I haven't thought about that one yet. But at the very least, there is a path forward for day/night only if you want to split that out into a separate PR.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
Developer-Butters: it reads easier to me. will remove the extra spaces.

SerialPrograms/Source/PokemonLZA/Programs/ShinyHunting/PokemonLZA_ShinyHunt_BenchSit.cpp
Developer-Butters: it corrects for any accidental movement on the way to the bench.  I can remove it here and elsewhere.

Developer-Butters:
> Now at least the day/night detection works. (I haven't checked the weather yet.)
> 
> But my main concern now is how long it takes to open/check the map. You're adding 6 seconds to open the map. This is done on every day night cycle. So for every day/night pair, you add 12 seconds of overhead. This means that any run duration less than 6 seconds will actually be slower by filtering. The majority of the use cases for walking forward are less than that.
> 
> IOW, the day/night detection needs to be a lot faster for this to be useful. Weather is a different story though. I haven't run the numbers on that to see whether it makes sense.
> 
> If the map detection can't be sped up, you'll need do something else. For day/night, it's easy since they always alternate between day/night. Do the detection at the beginning and remember what the parity it. Then only recheck maybe every 100 bench sits to make sure you didn't miss one and end up on the wrong parity.
> 
> For weather detection, that one is tricker. I haven't thought about that one yet. But at the very least, there is a path forward for day/night only if you want to split that out into a separate PR.

I've restored the comments that were erroneously removed locally. I've also updated the formatting locally. I hear you on the added time from the map open, and I'm adding a user defined periodic check similar to your periodic save. I'll run a test of it.

I also hear you on the weather thing. I was thinking it would be useful for pokemon like Goomy. But I think you're right the additional time it adds and the fact there are other, more effective ways available to shiny hunt these pokemon may not make the addition worth it. I have two thoughts: I can remove it entirely or I can add a disclaimer that this will greatly increase the amount of time it takes for you to use the program. I lean towards the former because it's such an edge case it probably would never be used.

Mysticial:
Then let's remove the weather part for now and clean up the day/night portion so that it checks only periodically. Then we we're good to go.

Developer-Butters:
I should have a new push tomorrow with the removal and all the current issues addressed.

Developer-Butters:
image

The new commit now caches the time and periodically checks by a user defined field. the overlay looks like the screenshot when running(I set the check to 3 sits for testing). I also decreased the time substantially in the check so it should be under 2000ms to check the map and continue running the program, so now it at most doubles the shortest run forward on the wiki page, which I think is the 2000 ms run in zone 9.

PR #1182, opened 2026-04-16
Even more RngHelper fixes

theastrogoth:
A few small changes:
- Fixed starter summary checking (there is not yet a Pokedex in the start menu. This must have gotten mixed up at some point during reorganization)
- added Summary detection 
- added a User Profile option (mostly for Switch 2, which doesn't remember the most recent profile when resetting)


SerialPrograms/Source/PokemonFRLG/Programs/ShinyHunting/PokemonFRLG_RngHelper.cpp
```diff
@@ -64,8 +65,15 @@ std::unique_ptr<StatsTracker> RngHelper_Descriptor::make_stats() const{
 }
 
 RngHelper::RngHelper()
-    : TARGET(
-        "<b>Target:</b><br>",
+    : PROFILE(
+        "<b>User Profile Position:</b><br>"
+        "The position, from left to right, of the Switch profile with the FRLG save you'd like to use.<br>"
+        "If this is set to 0, Switch 1 defaults to the last-used profile, while Switch 2 defaults to the first profile (position 1)",
```
Mysticial, line 71: This isn't strictly necessary since we already require everyone using Switch 2 to move their profile to the first slot if the game needs to SR.

If you look at the wiki for every program that needs to SR, it's one of the first requirements.

PR #1199, opened 2026-04-20
Azure, MacOS, Ubuntu improvements

Koi-3088:
- Extract the `Resources` folder for all OS.
- Use a .dmg installer for MacOS.
- Implement gstreamer for Ubuntu.
- Improve Azure build times for all OS (especially MacOS by consolidating building and notarization steps).
- Fix double-zip issue for Azure artifacts and make them more convenient.
- Add ability to run multiple instances on MacOS via terminal by running `open -n /Applications/SerialPrograms/SerialPrograms.app --args --profile <some profile name>`


Koi-3088:
Azure app bundle (Ubuntu AppImage, MacOS .app/.dmg) and Windows build artifacts work fine on my end. Local build works fine as well. But should be tested/verified since some build environment variables might be quirky/weird, and I don't want to break local builds or introduce regressions.

Mysticial (approved):
Holy shit this resource stuff is getting messy. lol

pifopi (approved):

Koi-3088:
> Holy shit this resource stuff is getting messy. lol

Yeah, a little bit. Which is why I wanted to just rip the bandaid off for application bundles while jw is dealing with the brunt of it.

.azure-pipelines/templates/macos-build.yml
```diff
@@ -103,92 +105,480 @@ stages:
...
+          set -e
+          DMG_NAME="SerialPrograms-MacOS-$(compiler)-$(architecture).dmg"
+          ZIP_NAME="SerialPrograms-MacOS-$(compiler)-$(architecture).zip"
+          CACHE_DIR="$(Pipeline.Workspace)/Arduino-Source-Internal/cache-notarized"
+          BUILD_TYPE="${{ parameters.buildType }}"
+
+          cd "$CACHE_DIR"
+          if [ "$BUILD_TYPE" = "PrivateBeta" ]; then
+              echo "=== Creating password-protected archive for PrivateBeta ==="
+              7z a -tzip -mx=9 "-p$ARTIFACT_PASSWORD" -mem=AES256 "$ZIP_NAME" "$DMG_NAME"
```
Ericzklm: I have an issue extracting the zip with the default MacOS Archive Utility, had to install third party tools to do so. Archive Utility supposedly doesn't support AES256. Archive Utility doesn't tell you what failed exactly so users likely assume the password is incorrect

Ericzklm (changes requested):

.azure-pipelines/templates/macos-build.yml
Koi-3088: Can just require 7zip for private beta for all OS, since all are password-protected in the same way. It's also trivial to install and has powerful functionality. What would otherwise be the proposed change that preserves the behavior?

Koi-3088: Also note that this is not new to this PR, the 7zip step was just consolidated into the building stage with the introduction of the .dmg file (was previously in the `macos-notarization.yml` which was removed).

EDIT: We do have a pinned message in the Discord beta channel that mentions the 7zip requirement.

Ericzklm: I wasn't aware that 7zip started supporting MacOS. It is only available as CLI though which might be non-intuitive for non-technical users. Given that it's just for the beta and that the command to run can just be given, it's likely not a big deal. Users can also just install Unarchiver instead.

The proposed change to support Archive Utility would be to use a different format when encrypting with 7zip or use Apple's `aea` to keep AES

Koi-3088: Did some reading. Seems like if I were to use Apple's `aea` I would have to do it via terminal if I want a custom password, but then users would also have to use terminal to extract, which kind of defeats the purpose.

Could remove `-mem=AES256` to use MacOS default encryption (should be ZipCrypto?) or explicitly specify it, not sure if there's a difference. Probably latter is better.

I'll test it locally using p7zip and will update the PR if Archive Utility looks happy. AES256 is probably overkill anyways.

Koi-3088: Seems to work on my end, pushed the requested change.

Ericzklm (approved):

PR #1205, opened 2026-04-22
FRLG Starter Auto-RNG

theastrogoth:
Automatically calibrates timings to hit a specific seed and advance for picking up a starter.

User setup (all easily copyable from ten-lines):
- Seed info
   - seed hex string
   - ordered list of nearby seeds (±5 works well)
   - delay for the target seed
   - button press for the target seed
- Target advances (total)
- select desired starter from dropdown
- save in front of starter's ball (on the south side of the table)

Works for both shiny and non-shiny targets. Tested several times with non-shiny targets (SID-finding has been slower than expected).


SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_RngDisplays.cpp
```diff
@@ -0,0 +1,185 @@
...
+void PossibleHitsDisplay::set(const std::vector<AdvRngState>& rng_states){
+    hits.set(get_hits_string(rng_states));
+}
+void PossibleHitsDisplay::set(std::map<AdvRngState, AdvPokemonResult>& hits_map){
+    std::vector<AdvRngState> rng_states = get_rng_states_from_map(hits_map);
+    set(rng_states);
+}
+
+void PossibleHitsDisplay::reset(){
+    hits.set("-");
```
Mysticial, line 183: Here you have `-` as the empty state. But `get_hits_string()` returns empty string on empty state. Which one should it be?

SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_StarterRng.h
```diff
@@ -0,0 +1,152 @@
...
+        const uint16_t& MAX_HISTORY_LENGTH,
+        double& SEED_CALIBRATION_FRAMES,
+        double& ADVANCES_CALIBRATION,
+        double& CONTINUE_SCREEN_ADJUSTMENT,
+        std::map<AdvRngState, AdvPokemonResult>& search_hits,
+        bool force_finish = false
+    );
+
+    bool walk_to_rival_battle(SingleSwitchProgramEnvironment& env, ProControllerContext& context);
+    bool auto_battle_rival(
```
Mysticial, line 105: The purpose of battling rival is to get lvl6 stats to refine the search?

theastrogoth: Yup. There are 3 ways to try to figure out what advance you're hitting when there are multiple possibilities (which is usually the case):
- Level up and use the new stats to get tighter IV ranges. This can be done with the rival battle and/or some Route 1 wild battles
- Keep RNG advances the same while slightly changing the seed delay to avoid hitting the exact same stats. Look for common advances (within ±2) in the search results across multiple resets
- Pick the search result that is closest to your target advances. This is only a good idea after some calibration has been done

Currently, the program attempts to use a combination of these. Things still work fairly well even if we leave out battles entirely -- I originally messed up the part that updates the search filters, so level-up stats weren't actually narrowing the search results, but the program still hit the target reasonably quickly.

SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_RngDisplays.cpp
theastrogoth: I've added a more descriptive message in the case where there are no search hits. `-` will correspond with the empty state before a search has been done, while `No matches found` will be displayed if a search turns up empty.

SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_RngDisplays.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+PossibleHitsDisplay::PossibleHitsDisplay()
+    : GroupOption("Possible Hits", LockMode::READ_ONLY)
+    , hits(false, "<b>Seeds/Advances:</b>", LockMode::READ_ONLY, "-", "")
+{
+    PA_ADD_STATIC(hits);
+}
+
+std::vector<AdvRngState> PossibleHitsDisplay::get_rng_states_from_map(std::map<AdvRngState,AdvPokemonResult>& hits_map){
```
Mysticial: Since you're not modifying the map, make it const.

PR #1206, opened 2026-04-23
Home living dex

Dalton-V:
New program for sorting a living dex.

This shares a lot of code with the normal box sorter. They could probably get combined in the future. Currently the setup for the user is too different. Hopeful that a separate program makes them read the wiki

Depending on the json should cover most of https://github.com/PokemonAutomation/ComputerControl/issues/210

Tasks remaining:

- [x] Shiny Dex Versions
- [x] OT Filter
- [x] Dex Json
- [x] File Output
- [x] Dry Run
- [x] Dex Options 
- [x] Dependent on https://github.com/PokemonAutomation/Arduino-Source/pull/1214 Due to false positives for Gmax/Tera


SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxNavigation.cpp
```diff
@@ -0,0 +1,402 @@
...
+#include "PokemonHome_BoxNavigation.h"
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonHome{
+using namespace Pokemon;
+
+
+// Move the red cursor to the first slot of the box
+// If the cursor is not add the first slot, move the cursor to the left and up one row at a time until it is at the first slot.
```
pifopi: Typo at

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxNavigation.cpp
```diff
@@ -0,0 +1,402 @@
...
+    const BoxCursor& cur_cursor,
+    const BoxCursor& dest_cursor,
+    Milliseconds GAME_DELAY
+) {
+
+    std::ostringstream ss;
+    ss << "Moving cursor from " << cur_cursor << " to " << dest_cursor;
+    env.console.log(ss.str());
+
+    // TODO: shortest path movement though pages, boxes
```
pifopi: Typo through

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxNavigation.cpp
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    //env.update_stats();
+    env.log(ss.str());
+    env.add_overlay_log("Empty: " + std::to_string(num_empty_slots) + "/30");
+
+    return first_pokemon_slot;
+}
+
+// Read the current summary screen and assign various pokemon info into `cur_pokemon-info`
```
pifopi: Typo - instead of _

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorterLivingDex.h
```diff
@@ -0,0 +1,102 @@
...
+    SimpleIntegerOption<uint16_t> LIVING_DEX_START_BOX;
+    SimpleIntegerOption<uint16_t> REJECT_BOX_START;
+    SimpleIntegerOption<uint16_t> REJECT_BOX_END;
+
+    // Living dex style.
+    EnumDropdownOption<LivingDexStyle> LIVING_DEX_STYLE;
+
+    // Dex filters
+    BooleanCheckBoxOption SHINY_DEX;
+    OCR::LanguageOCROption OT_NAME_LANGUAGE;
```
pifopi, line 88: IIRC, the reason why this wasn't already added is because OT name can be any language (on a per pokemon basis). If you traded, etc for pokemon with nickname, you're gonna trip that reader :/

pifopi (commented):
Quickly went through it ! Thanks for your PR

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorterLivingDex.h
Dalton-V: It can be any language however, a lot of people who collect a living dex eventually want every pokemon to be caught themselves. So all the desired pokemon would have the same OT name in the users language. If a pokemon has a OT name in a different language than the user the reader will return an empty string or a bunch of random chars. Either way they will get rejected and sent to the reject boxes.

SerialPrograms/cmake/SourceFiles.cmake
```diff
@@ -1573,8 +1573,12 @@ file(GLOB LIBRARY_SOURCES
     Source/PokemonHome/PokemonHome_Panels.h
     Source/PokemonHome/PokemonHome_Settings.cpp
     Source/PokemonHome/PokemonHome_Settings.h
+    Source/PokemonHome/Programs/PokemonHome_BoxNavigation.cpp
+    Source/PokemonHome/Programs/PokemonHome_BoxNavigatino.h
```
Mysticial: typo

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxSorterLivingDex.cpp
```diff
@@ -0,0 +1,598 @@
...
+        "<b>OT Name Filter:</b><br>Only keep pokemon with this OT name in the living dex. "
+        "Leave blank to accept any OT name.",
+        LockMode::LOCK_WHILE_RUNNING,
+        "",
+        "Enter OT name here..."
+    )
+    , VIDEO_DELAY(
+          "<b>Capture Card Delay:</b>",
+          LockMode::LOCK_WHILE_RUNNING,
+          "400 ms"
```
Mysticial, line 137: Why do you need this option? No other program has ever needed it.

SerialPrograms/Source/PokemonHome/Programs/PokemonHome_BoxNavigation.cpp
```diff
@@ -0,0 +1,402 @@
...
+namespace PokemonHome{
+using namespace Pokemon;
+
+
+// Move the red cursor to the first slot of the box
+// If the cursor is not at the first slot, move the cursor to the left and up one row at a time until it is at the first slot. 
+bool go_to_first_slot(
+    SingleSwitchProgramEnvironment& env,
+    ProControllerContext& context,
+    Milliseconds VIDEO_DELAY
```
Mysticial, line 46: Oh ugh... this is where you use it. I guess this is just carry over what from the original box sorter had.

I don't block the PR here, but what I'd do is:
1. Choose the target location.
2. Read your current location.
3. Blindly move from current to target using fast button presses.
4. Use inference to detect when the cursor reaches the destination.

The blind sequence in #3 would be wrapped with the `run_until()` in #4. And if you fail to reach the target after the moves are complete + a long grace period (like 5 seconds?), go back to #2 and try again. Thus a self-correcting algorithm that will run as fast as possible without caring about video delay.

PR #1214, opened 2026-04-26
Added proper detection for gmax and tera type

Dalton-V:
Resolves the following issue:
TODO: gmax symbol is at the same location as Tera type symbol! Need better detection to tell apart
Tera types were causing false positives for Gmax. Breaking the living dex sorter.

Added template matching for Gigantamax
Added call to TypeReader for Tera types

Support sorting by Tera type

- [x] Test Gmax Detection
- [x] Test Tera Detection

### DIFF
--- a/SerialPrograms/Source/Pokemon/Inference/Pokemon_TypeReader.cpp
+++ b/SerialPrograms/Source/Pokemon/Inference/Pokemon_TypeReader.cpp
@@ -96,24 +96,6 @@ struct Gen9TypeSpriteDatabase {
     }
 };
 
-struct TeraTypeSpriteDatabase {
-    std::map<PokemonTeraType, TypeSprite> m_type_map;
-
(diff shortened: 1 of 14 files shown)


Mysticial:
Is this stuff specific to Home? Or in general? Since a lot of the type detector logic is copy-pasted from the Sword/Shield one? So if this is generic (the type symbols are the same), we can merge that. If it's not and these are specific to Home, then you should move these into the Pokemon Home directory.

Dalton-V:
I put it in the same class as the detection logic was the same, just using different sprite resources. These icons are very similar to the ones on the map in SV. However, after some testing they appear to be different enough to cause miss-reads.
 
Moved tera detection to home inference folder. Updated the resource location as these are Home specific. https://github.com/PokemonAutomation/Packages/pull/57

PR #1216, opened 2026-04-27
FRLG Gift Auto-RNG

theastrogoth:
This adds a program that automatically calibrates timings to hit a specific seed and advance for picking up a gift or Game Corner prize.

The in-game setup is the same as the RNG Helper, and the program options are the same as the Starter RNG program, with two additional options:
- a maximum number of rare candies to use for leveling up to get tighter IV ranges
   -  in-game setup: these need to be at the top of the bag 
   - *OR* this can just be set to 0 in the program
- a checkbox to toggle Teachy TV use, similar to the RNG Helper

Works for both shiny and non-shiny targets. Tested at the Game Corner with the Teachy TV


SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_HardReset.cpp
```diff
@@ -78,7 +78,8 @@ void rng_start_game_from_home(
         int ret1 = wait_until(
             console, context,
             std::chrono::seconds(2),
-            { black_screen }
+            { black_screen },
+            1ms
```
Mysticial: Didn't you say you were doing to move off of using inference for this? Or is that coming in a later PR?

theastrogoth: I went ahead and added it here. It's a rather ugly solution (splitting things up into 4 steps rather than handling it with a single state machine), and there are probably some irrelevant detectors present in some steps since I'm not sure about the sequencing, but I think it should be able to handle popups.

Seed consistency is *much* better.

PR #1218, opened 2026-04-28
FRLG Static Encounter Auto-RNG

theastrogoth:
This adds a program that automatically calibrates timings to hit a specific seed and advance for static encounters (covering similar hunts to the Legendary Reset program). 

This is very similar to the Gift RNG program -- it just adds a basic autocatcher function. 

A couple of things added/changed in addition to the program itself:
- fixed an issue with RNG resets that would probably have led to unintentionally accepting the update popup
- added a `!is_white` check in the `BattleMenuDetector`, which was being triggered by white screens
- added a `DexRegistrationDetector` for help with reliably exiting the battle after a catch


Mysticial:
Mostly rubber stamping these since you own all of it anyway. :)

PR #1223, opened 2026-05-02
FRLG Wild Encounter Auto-RNG

theastrogoth:
This adds an RNG manipulation program for wild encounters (grass, fishing, etc).

Relies on a few additional resource files (https://github.com/PokemonAutomation/Packages/pull/58)

This brings the auto-RNG programs up to coverage of all obtainable Pokemon in FRLG except for the roaming legendaries.

Changes to other RNG programs
- small tweaks to how possible hits are stored
- changed several errors related to user error to avoid sending reports

There are a few small changes to the FRLG summary reader as well:
- modified HP box size
- tweaked how HP reading is handled (it had been failing for some 3 digit values)
- added the optional use of a subset of Pokemon names for species OCR


SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_WildRng.cpp
```diff
@@ -0,0 +1,1016 @@
...
+            {EncounterType::goodrod, "goodrod", "Good Rod"},
+            {EncounterType::superrod, "superrod", "Super Rod"},
+        },
+        LockMode::LOCK_WHILE_RUNNING,
+        EncounterType::grass
+    )
+    , GAME_LOCATION(
+        "<b>Location:</b>",
+        {
+            {GameLocation::altering_cave, "altering_cave", "Altering Cave"},
```
Mysticial, line 117: I'll merge this as-is since I don't want to block it, but this list is so long, it might be a good idea to use a `StringSelectOption` instead. It supports typing in to filter results. Look at how the Pokemon selector uses it.

You can even add your own icons to it if you want.

PR #1227, opened 2026-05-03
update show_update_nag()

jw098:
Updates the logic to cover for the case where your current version is behind both the public and private beta version.
In this case, "Skip this version" doesn't work since the saved version to skip will alternate between the public and private beta version number. e.g. If your current version number is 0.63.7, the pop-up will alternate between 0.63.8 (the beta) and 0.63.9 (public release).


jw098:
I missed the fact that skip_version is a string, and not a number that can be compared.
So, fixing this requires more effort than I anticipated.

PR #1228, opened 2026-05-04
SV Egg Autonomous: Added the option with backup saves

Boombaastical:
My purpose was to be able to take a video of the trainer profile with the shiny hatching from the egg in a single video. In our current option, my best bet was to:
-Save after picnic and all eggs have been fetched
-Save before every batch of 4-5 eggs

In both cases, it would use up ingredients, and it takes more time because the program needs to release all pokemon at the end.

So instead, I added an option to create a backup save after fetching all 30 eggs, but keeping the logic of "Save at the beginning and after keeping a baby". This way, it works exactly the same, but the backup save allows to go back to the moment where the shiny was part of the eggs, and thus I can manually take a video of it.

Backup saves in SV:
-Are created by:
--healing pokemon at a pokemon center (the option I chose for this PR)
--after finishing an online trade

-Are not created by:
--Any operation done by the current SV Egg Autonomous
--Flying somewhere, close or far
--Moving to another area

I have already tested the feature, and I could record the 3 videos in attachment.

https://github.com/user-attachments/assets/415e52f1-96d6-4f8d-8ce7-b5a09825d8ee

https://github.com/user-attachments/assets/ba2a5fc8-2e66-4a64-80dd-0d26e963d458

https://github.com/user-attachments/assets/b0dcdb75-3258-4158-aa5c-aa1431ddf607

### DIFF
--- a/SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
+++ b/SerialPrograms/Source/PokemonSV/Programs/Eggs/PokemonSV_EggAutonomous.cpp
@@ -15,7 +15,10 @@
 #include "CommonFramework/ProgramStats/StatsTracking.h"
 #include "CommonFramework/VideoPipeline/VideoFeed.h"
 #include "CommonTools/StartupChecks/VideoResolutionCheck.h"
+#include "CommonTools/Async/InferenceRoutines.h"
 #include "NintendoSwitch/Commands/NintendoSwitch_Commands_PushButtons.h"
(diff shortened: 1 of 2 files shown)


Mysticial:
This is overly complicated for what you want to do. Just use audio to detect the shiny sound that's played during the hatch. Then take a video via the screenshot button after the egg is done hatching.

Boombaastical:
The point is also to record the Trainer ID. For this, I don't want to add this extra step as part of the program, as this is specific for me, but the fact that you can rehatch eggs at no additional cost is an improvement over the other saving options.

Mysticial:
Closing due to high risk and little benefit.

PR #1237, opened 2026-05-10
ZA Stats Reset: Add Hoopa

ZHDreamer:
I had been waiting a long time for Hoopa reset, so I decided to implement it myself. Hope this is not offensive.

Tested on Switch OLED with PABotBase2 for ~100 rounds with 1 recoverable error and no unrecoverable error. I don't have Switch 2, so I can't test on it.

Here is some instruction to run the program:

Preparation for Hoopa:
1. Start Side Mission 196 until the portal appears.
2. Have a Level Boost (100+) donut ready (alternatively, Attack Power or Bug-type Move Power). Configure Donut Down Scrolls in settings so it scrolls to the correct donut.
3. Make room on the top left slot in the box system, make sure that this is where the cursor is when opening the box system from the main menu.
4. Select Use Plus Moves in Battle AI.
5. Lead with a Level 100 Heracross (252+ Atk, Life Orb) with Megahorn on top slot (button X). Hoopa Unbound is 4× weak to Bug type and has low Def, so with right donut and plus move, a one-hit KO is guaranteed. The program can still execute multiple moves if needed.
6. Start the program in game.

### DIFF
--- a/SerialPrograms/Source/PokemonLZA/Programs/NonShinyHunting/PokemonLZA_StatsReset.cpp
+++ b/SerialPrograms/Source/PokemonLZA/Programs/NonShinyHunting/PokemonLZA_StatsReset.cpp
@@ -458,7 +458,7 @@ void StatsReset::program(SingleSwitchProgramEnvironment& env, ProControllerConte
             pbf_move_left_joystick(context, {0.05, 1}, 4800ms, 0ms);
             pbf_mash_button(context, BUTTON_A, 1s);
             context.wait_for_all_requests();
-            wait_until_overworld(env.console, context);
+            wait_until_overworld(env.console, context, 10s);
 
             // run to the right of roof
             pbf_press_button(context, BUTTON_Y, 100ms, 1s);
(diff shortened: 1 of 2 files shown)


ZHDreamer:
Maybe we should add take a day/night skip before running the program to the instructions, since there is currently no day/night change handler. This is especially important for the Hoopa reset because of the long overworld routing.

NympheaR:
@ZHDreamer Agree on the day/night handler since most of the other resets have short routes and can self-correct safely. Is that something you want to get to or should I work on it once I have a bit more time on my hands?

SerialPrograms/Source/PokemonLZA/Programs/NonShinyHunting/PokemonLZA_StatsReset.cpp
```diff
@@ -441,6 +447,122 @@ void StatsReset::program(SingleSwitchProgramEnvironment& env, ProControllerConte
...
+                go_home(env.console, context);
+                reset_game_from_home(env, env.console, context, true);
+                continue;
+            }
+            context.wait_for(100ms);
+            // run toward hoopa to catch
+            ssf_press_button(context, BUTTON_B, 0ms, 2s, 0ms);
+            pbf_move_left_joystick(context, {0.05, +1}, 1500ms, 0ms);
+            pbf_mash_button(context, BUTTON_A, 1s);
+            // incase did not run
```
NympheaR: I might update it for the Magearna reset at some point but run_towards_gate_with_A_button(); after a direction align might be better here as an existing detection-based alternative to trying movement again.

ZHDreamer:
I have two ideas for handling the day/night transition.

1. Detect day/night changes during the overworld loop and restart the main loop from the beginning. The fly sequence at the beginning will create a new backup save after the day/night change. This is a more general solution and could be reused for other resets as well.

2. Start from the Bleu Pokémon Center instead of Café Soleil. If we make the backup save there, we can reload and run without flying. Since there’s no flying, time won’t advance and day/night changes should never happen (unless the program is started too close to a day/night transition). This is simpler but Hoopa specific.

If we go with the first approach, it may be worth adding it to other reset programs as well.

> Is that something you want to get to or should I work on it once I have a bit more time on my hands?

I can probably work on it since I have a lot of free time right now. I’ll just need some guidance on detecting day/night changes properly if we go with the first idea. Would it be possible for me to join the developer corner in Discord? It would make it easier for me to ask questions while working on this.

SerialPrograms/Source/PokemonLZA/Programs/NonShinyHunting/PokemonLZA_StatsReset.cpp
ZHDreamer: Why is a direction align needed before `run_towards_gate_with_A_button()` instead of just running directly in that direction?

NympheaR: sorry bad wording whether or not the realign is needed depends on the angle of approach, there is a chance you might overshoot the NPC (since the A button detection requires it to be on screen for a certain amount of time) so it's worth doing some testing. I don't have a hoopa ready savefile ready and my cfw switch is still in the repair shop so if you can test to see if the realign can be omitted that'll be great.

ZHDreamer: I'll test it

ZHDreamer:
Changed to running from Bleu Pokémon Center. Tested for a night without any problems.

NympheaR (approved):
tested on my end, should be good to go

PR #1251, opened 2026-05-20
Another round of FRLG RNG updates

theastrogoth:
This makes a few changes based on some feedback from testers:
- adds a field that displays the stats of the user's RNG target to help catch when there are mistakes in the inputs
- adds additional fields to display read species and level to make OCR issues more obvious
- more details about user settings are logged to make debugging easier
- tweaks the layout of some of the RNG program inputs to have a more coherent ordering
- change some errors to `UserSetupError` where appropriate

... and also addresses a bug when trying to do wild RNG in Safari Zone West (reported by Sheep Punter)


SerialPrograms/Source/Pokemon/Pokemon_AdvRng.cpp
```diff
@@ -664,11 +650,11 @@ AdvRngFilters observation_to_filters(const AdvObservedPokemon& observation, cons
 std::string gender_to_string(const AdvGender& gender){
     switch (gender){
     case AdvGender::Male:
-        return "Male";
+        return "♂";
     case AdvGender::Female:
-        return "Female";
+        return "♀";
```
Mysticial: It'd generally not a good idea to put non-ascii characters in source files. Can you use an escape character? You can look in `Pokemon_Strings.cpp`.

SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_BlindNavigation.cpp
```diff
@@ -779,6 +779,7 @@ void perform_blind_sequence(
         modified_ingame_delay = timings.ingame_delay - 51430;
         walk_to_safarizonewest(context);
         use_sweet_scent(context, modified_ingame_delay, true);
+        return;
```
Mysticial, line 782: lol

PR #1253, opened 2026-05-21
Better reliability for FRLG auto-catching and species detection

theastrogoth:
This adds
- Sprite matching for FRLG. Currently only used as a fallback for the summary reader if there are no good results from OCR or for double-checking Nidoran's gender
- Sound detection for the fanfare that plays on a successful catch
- Better handling of evolution after leveling up with Rare Candy

... all useful for making the FRLG Wild RNG program more reliable, particularly in the Safari Zone

Depends on https://github.com/PokemonAutomation/Packages/pull/66

EDIT: also added a fix for the seed calibration bug reported by Sollisnexus

### DIFF
--- a/SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_PokemonSpriteReader.cpp
+++ b/SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_PokemonSpriteReader.cpp
@@ -1,0 +1,98 @@
+/*  Pokemon FRLG Sprite Reader
+ *
+ *  From: https://github.com/PokemonAutomation/
+ *
+ */
+
+#include "CommonFramework/ImageTypes/ImageViewRGB32.h"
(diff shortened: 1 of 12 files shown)


theastrogoth:
All CI checks passed on one run or the other. 

The first time, Windows checks failed ("The hosted runner lost communication with the server"). The second time ran into an issue with Ubuntu dependencies. 

Happy to close and reopen this PR again to triple check if needed, but I figured it might be better to avoid it if it's not necessary.

pifopi:
You can rerun jobs at will ! Though GitHub seems to have issues lately with their runners. I know they're working on it. We'll see how it goes

theastrogoth:
I might be dumb and overlooking the place to do this in the Github web interface, but I think that rerunning CI jobs might be locked behind a particular permission/role.

Anyways, thanks for restarting the failed jobs! Looks like they got through the Ubuntu dependencies this time 🎉

PR #1254, opened 2026-05-21
Pokopia Auto-Miner

Ericzklm:
Automatically mine a Dream Island to collect Pokémetal, Rare Pokémetal, iron ore, gold ore

### DIFF
--- a/SerialPrograms/Source/PokemonPokopia/Inference/PokemonPokopia_MovesDetection.cpp
+++ b/SerialPrograms/Source/PokemonPokopia/Inference/PokemonPokopia_MovesDetection.cpp
@@ -6,8 +6,11 @@
 
 #include "Common/Cpp/Exceptions.h"
 #include "Kernels/Waterfill/Kernels_Waterfill_Types.h"
+#include "CommonFramework/Globals.h"
 #include "CommonTools/ImageMatch/WaterfillTemplateMatcher.h"
 #include "CommonTools/Images/WaterfillUtilities.h"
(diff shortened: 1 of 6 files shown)


Ericzklm:
Closing this feature. This would be a fairly large and complex program and there's not enough demand or developer bandwidth for it at the moment.

In case anyone is interested in picking this up, here's a braindump of what I had in mind
Concept: Use Potato Hamburger Steaks and Rollout on a dream island to layer mine a square for metal ores and collect them in the inventory
High-level idea:
- User digs a hole on the dream island in the structure behind spawn since that's in a predictable location
- User empties inventory except for Potato Hamburger Steaks in the first few slots
- Program starts by using Pokedex menu to travel back to Drifloon to reset position and camera
- Use Camera app to lock camera direction towards the Island
- Use the grid lock on movement (left bumper?) to navigate into the hole
- Rollout in a straight line forward for a user-defined amount of time for a 2-wide tunnel
- Repeat travel/camera/positioning routine to fall back into the hole
- Do collection routine down the tunnel using the suck
- Travel back to Drifloon, sort inventory, drop all junk except for desired items
- Redo camera/positioning routines to navigate into the hole again
- Dig two blocks right, then loop the sequence: Rollout, travel/position, collect, travel, inventory management, position
- Watcher on the PP bar and top up with a Potato Hamburger Steak when a user-defined threshold is reached
- Stop when no more food is in the inventory
- This flow should ensure you never go "backwards" which would be a problem because the state of the world behind your actions is unknown and can lead to getting stuck

PR #1259, opened 2026-05-23
First fix to the 'Cannot open Rotom phone menu after multiple retries...' error

Boombaastical:
Since it was currently stuck on a reproducible loop with this error, I could just test it multiple times. I simply added a EggBlackDialogBoxDetector, as sneakily enough, the BlackDialogBoxDetector was just not aligned with the dialog box when the eggs hatch (pictures attached, they're slightly different).

Tested it for some time now, it got out of the loop, I will leave it as a draft for now and then update it to a proper PR if it continues to loop without that error. The other error of having an egg left in the party I believe isn't that big of a deal as every time it happens, the reload loop fixes it.
BlackDialogBox1
BlackDialogBox2


Boombaastical:
Will continue to run it to see if it's fixed

PR #1265, opened 2026-05-24
Even more FRLG RNG fixes

theastrogoth:
Several small fixes:
- add missing settings entry for Fanfare detection threshold
- fix uninitialized calibrations variable
- store updates to `launch_delay` to avoid needing to rediscover them during each reset, and use smaller increments to avoid possible timing issues
- skip through dialogue after Hypno battle before checking summary
- remove redundant logging of seed values


theastrogoth:
One of the changes here duplicates #1264 (apologies, didn't notice it until now)

PR #1272, opened 2026-05-29
Resource Downloading

jw098:

PR #1273, opened 2026-05-29
Add required resources to program descriptors

jw098:
From the large PR #1272.


SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceType.h
```diff
@@ -0,0 +1,19 @@
...
+
+#ifndef PokemonAutomation_ResourceType_H
+#define PokemonAutomation_ResourceType_H
+
+
+namespace PokemonAutomation{
+
+enum class ResourceType {
+    OCR,
+    YOLO_MODEL_AREA_ZERO,
```
Mysticial: I'm not sure if we want an enum here vs. a slug system. In the long term, I imagine this will get really long. Let me think about this.

PR #1274, opened 2026-05-30
RequiredDownloadManager

jw098:


SerialPrograms/Source/CommonFramework/ResourceDownload/RequiredDownloadManager.h
```diff
@@ -0,0 +1,97 @@
...
+    // and so can't be moved unless it's a pointer. 
+    // Objects can't be added to a vector if they can't be moved/copied.
+    // also, it being a pointer allows it to be forward declared. 
+    // this is needed since RequiredDownload also holds a reference to this Session
+    std::vector<std::shared_ptr<RequiredDownload>> m_required_downloads;  
+
+    // queue of downloads
+    // each download is represented by an int, 
+    // which is their index within the local resource list. 
+    std::vector<uint16_t> m_download_queue;
```
Mysticial: This looks dangerous? Can the list change? Also, use `size_t` for stuff like array indices.
I feel like you should be using the slug instead of an index.

SerialPrograms/Source/CommonFramework/ResourceDownload/DownloadThread.cpp
```diff
@@ -0,0 +1,173 @@
...
+void DownloadThread::start_download_thread(){
+    m_worker = GlobalThreadPools::unlimited_normal().dispatch_now_blocking(
+    [this]{
+
+        bool success = false;
+        {
+            std::unique_lock<Mutex> lock(m_download_lock);
+            m_download_cv.wait(lock, [this] { return m_hooks.is_ready_to_start() || m_stopping; });
+
+            if (m_stopping) return;
```
Mysticial: I don't see `m_stopping` ever being used outside this download thread.

So:
1. It doesn't need to be atomic.
2. Is it needed at all?

Since it's only ever read at the start of this function here and will always be false. Then it's only ever set when you leave this function. It looks like this is leftover from when you switched to using `Cancellable`.

SerialPrograms/Source/CommonFramework/ResourceDownload/DownloadThread.cpp
```diff
@@ -0,0 +1,173 @@
...
+        // runs when lambda is finished
+        // updates action state, removes self from download queue
+        struct ScopeGuard {
+            DownloadThread* thread_ptr;
+            bool& success_ref;
+            ~ScopeGuard() {
+                thread_ptr->m_stopping = true;
+                thread_ptr->m_hooks.on_finished(success_ref);
+            }
+        } guard{this, success};
```
Mysticial: Full this class out into its own file as `Common/Cpp/ScopeExit.h`.

https://github.com/PokemonAutomation/Arduino-Source/blob/main/SerialPrograms/Source/CommonFramework/Main.cpp#L63

Then do:
```
ScopeExit on_exit([&, this]{
    m_stopping = true;  //  Which I assume you can delete this if the variable is going away.
    m_hooks.on_finished(success);
});
```

SerialPrograms/Source/CommonFramework/ResourceDownload/RequiredDownloadManager.cpp
```diff
@@ -0,0 +1,173 @@
...
+            if (expected_version_num == remote_version_num){
+                // remote version matches what we expect
+                // add the resource to the download list
+                // m_required_downloads.emplace_back(RequiredDownload{remote_resource});
+                required_download_list.emplace_back(std::make_shared<RequiredDownload>(*this, remote_resource, m_queue_lock, m_cv));
+
+            }else if (expected_version_num < remote_version_num){
+                // remote version is more updated than we expect
+                // warn the user to upgrade CC.
+                // regardless, add the resource to the download list
```
Mysticial, line 82: Does downloading a newer version overwrite the existing one? I don't want this to become a forced update if the resource gets overwritten, then CC won't run because it expects the older version.

Mysticial (approved):

PR #1275, opened 2026-05-31
FRLG Roaming Legendary RNG

theastrogoth:
This adds an RNG program for Raikou/Suicune/Entei. It takes quite a bit longer than other RNG programs (I'm averaging around 2 encounters per hour so far) due to the lack of a reliable way to find roaming legendaries.

### In-game setup:
- Have the following items:
   + Rare Candy - top of bag
   + Max Repel - 2nd spot from top
   + Master Ball - top of bag
- have a Pokemon between Lv5 and Lv49 as your lead
   + having a Pokemon lv51 or higher will make it impossible to encounter the roaming legendary while Repel is active
- have a Pokemon that knows Fly in the last occupied spot in your party
- have at least one free spot in your party
- get the Sapphire and save to the left of Celio *before* giving it to him
   + it's important to be on the left so that the blind pathing works

### Other small changes made here:
- adds the `roaming` target and a language field (which is only used for the `roaming` target) to the RNG Helper program


SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_BlindNavigation.cpp
```diff
@@ -653,6 +672,18 @@ void check_timings(
             );
         }
         return;
+    case PokemonFRLG_RngTarget::raikou:
+    case PokemonFRLG_RngTarget::entei:
+    case PokemonFRLG_RngTarget::suicune:
+    case PokemonFRLG_RngTarget::roaming:
```
Mysticial, line 678: Why a separate case for roaming when you have the other 3?

Mysticial:
Can you separate out the small fixes into a separate PR and leave this one as roaming only?

SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_BlindNavigation.cpp
theastrogoth: That's for the RNG Helper, which doesn't ever check stats/IVs and doesn't need to know about the species. There's already a few similar cases for some other targets like starters (`bulbasaur`, `squirtle`, `charmander`, and `starters`), fossils, and stationary encounters, since they share button press sequences within each group.

It's mostly a consequence of making the RNG Helper program first. I can go through and replace the "general" cases with specific ones if that's cleaner (e.g. `starters` becomes `bulbasaur` when the exact species doesn't matter).

PR #1276, opened 2026-05-31
Resource Download widgets - popup dialog when starting program

jw098:


SerialPrograms/Source/CommonFramework/ProgramSession.cpp
```diff
@@ -34,6 +34,7 @@ ProgramSession::ProgramSession(const ProgramDescriptor& descriptor)
     , m_logger(global_logger_raw(), "Program")
     , m_timestamp(current_time())
     , m_state(ProgramState::STOPPED)
+    , m_download_manager(descriptor.required_resources())
```
Mysticial, line 37: This has me thinking. What if there are multiple programs at the same time and thus multiple `DownloadManager`s? Will they conflict? Is it possible for two instances to simultaneously try to download the same resource at the same time?

jw098:
Replaced the PR.

PR #1280, opened 2026-06-06
FRLG Egg RNG Program

theastrogoth:
This adds an RNG program for hatching eggs at the Four Island daycare.

It's a bit more complicated than the other RNG programs in a few ways:
- There are two different frames that need to be hit rather than one
   + This makes it necessary to save the game after (hopefully) hitting the first frame
   + I've tried to make it unlikely that the program saves and commits to a bad egg, but it's not totally guaranteed
- The target's IVs are influenced by the parents
- There are some frames that do not generate an egg, making calibration harder
- Ten Lines does not support searching for good frames for Egg RNG, and using PokeFinder is kind of a pain
   + I've mocked up a basic web app for searching for Held/Pickup frames, which might be helpful for eventual users: http://theastrogoth.github.io/frlg-egg-search

On the plus side, it makes it possible to obtain shiny Pokemon with *any* IVs and nature, regardless of your TID and SID, which isn't true of any Gift/Wild Pokemon.

There are also some small changes to things that are shared with other RNG programs. ~~When this is ready, I'll open a separate PR with the tweaks that affect shared code.~~ Done


theastrogoth:
Changes made in shared code are handled by #1283

theastrogoth:
After #1286, I think this is ready for a private beta!

SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_EggRng.cpp
```diff
@@ -0,0 +1,1443 @@
+/*  Egg RNG
```
Mysticial, line 1: This file is huge. If you want to break this up into smaller files, feel free to do that. Probably a lot of helper functions that can be extracted? Since I think we might have more egg routines like this in the future.

theastrogoth: I've tried to tidy things up a bit. Since this affected some other RNG programs, this has mostly been separated out into #1291 

This file is still big after the reusable parts have been elsewhere, but hopefully a little bit more readable

PR #1282, opened 2026-06-08
A few fixes for FRLG RNG programs

theastrogoth:
This fixes:
- **~~Review requested~~:**✅ Fixed detection issue with `BattlePokemonDetector`. This previously used `is_solid()`, but needed relatively narrow tolerance to avoid also being triggered by the background after fainting (solid white, or close to it). This was a bad idea, so I've switched to `is_grey()` to allow it to be used with more generous tolerances, as well as added negative locations for detecting backgrounds that would indicate fainting. Bug reported by... Leon, if I recall correctly? They changed their Discord name to "Feel free to ping me in reply" lol. 
   + also added solid-color detection for background behind HP bar area  
- **~~Review requested~~:**✅ Fixed issue with summary reading with a bright capture card in the FRLG `StatsReader`. I don't understand all of the choices in the FRLG OCR pipeline (not written by me), so I tried to leave it mostly intact. My solution was to filter/preprocess the lilac background based on its lower green value to make sure it's distinguished from the white text as well as add multi-filtering for name OCR. Bug reported by rageflapjack
- Fixed a bug preventing the RNG calibration from making use of resets with multiple remaining results if a program is run with 0 rare candies. 
- Added update for calibration history when a recoverable error happens


SerialPrograms/Source/PokemonFRLG/Inference/PokemonFRLG_BattlePokemonDetector.cpp
```diff
@@ -44,21 +44,21 @@ bool BattlePokemonDetector::detect(const ImageViewRGB32& screen){
...
     ImageViewRGB32 top_image = extract_box_reference(game_screen, m_top_box);
     ImageViewRGB32 bottom_image = extract_box_reference(game_screen, m_bottom_box);
-    if (is_solid(left_image, { 0.3418, 0.3418, 0.3164 })
-        && is_solid(right_image, { 0.2578, 0.3868, 0.3554 }, 0.075, 5)
-        && is_solid(top_image, { 0.3418, 0.3418, 0.3164 })
-        && is_solid(bottom_image, { 0.2578, 0.3868, 0.3554 }, 0.075, 5)
+    if (is_grey(left_image, 680, 770, 20)
+        && is_grey(right_image, 220, 320, 20)
+        && is_grey(top_image, 680, 770, 20)
+        && is_grey(bottom_image, 220, 320, 20)
```
Mysticial, line 53: Since you mention struggling here, can you just add more detection locations?

theastrogoth: I can't think of any good ones, unless we want to detect the sprites. Battle backgrounds change by location, so we're stuck with the Name/Level/HP bar area, I think

Mysticial: Have you tried negative locations? Since you know what the common false positives are, add boxes that will detect those false positives and require that they don't pass for this detection.

The Home screen detector as well as the BDSP dialog detector uses some negative detections to filter out some backgrounds.

theastrogoth: Added! I was a little hesitant to do this in case there were many background colors to handle depending on in-game location, but as far as I can tell, there are only two (brown in caves, white elsewhere)

PR #1294, opened 2026-06-19
Work around Discord's privileged intents restrictions

Koi-3088:
Work around the upcoming restrictions to Discord's privileged intents, implement a button to register slash commands, fix and improve readability of certain commands.

### DIFF
--- a/SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
+++ b/SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
@@ -29,8 +29,8 @@ Client& Client::instance(){
 
 
 Client::~Client(){
-    disconnect();
     GlobalSettings::instance().DISCORD->integration.register_slash_button.remove_listener(*this);
+    disconnect();
 }
 
 bool Client::is_initialized(){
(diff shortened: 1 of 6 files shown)


SerialPrograms/Source/Integrations/DppIntegration/DppClient.cpp
```diff
@@ -30,6 +30,7 @@ Client& Client::instance(){
 
 Client::~Client(){
     disconnect();
+    GlobalSettings::instance().DISCORD->integration.register_slash_button.remove_listener(*this);
```
Mysticial, line 32: Move this to start of destructor in case the button is clicked in the middle of the disconnect.

SerialPrograms/Source/Integrations/DppIntegration/DppClient.h
```diff
@@ -42,12 +43,26 @@ class Client : protected DppCommandHandler::Handler{
 
 private:
     void run(const std::string& token);
+    virtual void on_press() override{
+        std::lock_guard<std::mutex> lg(m_register_lock);
+        if (m_handler && m_handler->slash_commands_enabled && GlobalSettings::instance().DISCORD->integration.register_slash_button.is_enabled()){
```
Mysticial: You shouldn't need to check if the button is enabled since you can't click the button if it's disabled.

Mysticial (approved):

PR #1298, opened 2026-06-23
Resource Downloading

jw098:


jw098:
This PR is now ready to be reviewed. Let me know if you want the PR to be split up.

SerialPrograms/Source/NintendoSwitch/Framework/NintendoSwitch_SingleSwitchProgramSession.cpp
```diff
@@ -135,6 +135,24 @@ void SingleSwitchProgramSession::internal_stop_program(){
     }
 }
 void SingleSwitchProgramSession::internal_run_program(){
+    CancellableHolder<CancellableScope> scope;
+    {
+        std::lock_guard<Mutex> lg(program_lock());
+        if (current_state() != ProgramState::RUNNING){
+            return;
+        }
+        m_scope.store(&scope, std::memory_order_release);
```
Mysticial, line 144: I should really redesign this `m_scope` mechanism here. I hate. But that's for a future cleanup.

SerialPrograms/Source/NintendoSwitch/Framework/UI/NintendoSwitch_SingleSwitchProgramWidget.cpp
```diff
@@ -167,7 +176,29 @@ void SingleSwitchProgramWidget2::error(const std::string& message){
...
+
+    QMetaObject::invokeMethod(this, [message]{
+        QMessageBox box;
+        box.critical(nullptr, "Error", QString::fromStdString(message));
+    });
+    m_popup_is_open.store(false);
+}
 
+void SingleSwitchProgramWidget2::download_added(std::shared_ptr<ResourceDownload> download_ptr){
+    QMetaObject::invokeMethod(this, [this, download_ptr]{
```
Mysticial, line 192: I wonder if this works:
```
QMetaObject::invokeMethod(this, [this, download_ptr = std::move(download_ptr)]{
```
This moves it into the lambda instead of copying.

Same with the other places.

SerialPrograms/Source/NintendoSwitch/Framework/UI/NintendoSwitch_SingleSwitchProgramWidget.cpp
```diff
@@ -106,6 +107,10 @@ SingleSwitchProgramWidget2::SingleSwitchProgramWidget2(
     m_actions_bar = new RunnablePanelActionBar(*this, m_session.current_state());
     layout->addWidget(m_actions_bar);
 
+    m_downloads_table = new ProgramResourceDownloadTableWidget(*this);
```
Mysticial, line 110: I wonder if this should be done lazily (on first use). The vast majority of programs won't need this. So unconditionally instantiating its sounds wasteful especially since the widget is non-trivial.

Don't need to do anything about it in this PR since it's big enough. Maybe a future cleanup.

SerialPrograms/Source/CommonFramework/ResourceDownload/ProgramMissingResourceTracker.cpp
```diff
@@ -0,0 +1,164 @@
...
+    : m_scope(scope)
+    , m_promise(p)
+    , m_session(session)
+{
+    scope.add_cancel_listener(*this);
+}
+
+void ProgramMissingResourceTracker::add_resource(const std::shared_ptr<ResourceDownload>& resource){
+    m_missing_resources.insert(resource);
+    resource->add_listener(*this);
```
Mysticial, line 36: If `add_listener()` throw here, you'll be left in a bad state which I think will hang.

I suggest wrapping it in an exception and if it throws, remove the resource (undo the 1st line) and rethrow exception.

Mysticial (approved):

PR #1300, opened 2026-06-23
GlobalResourceDownloadManager

jw098:


jw098:
Note: RequiredDownload was renamed to ResourceDownload

SerialPrograms/Source/CommonFramework/ResourceDownload/ResourceDownload.h
Mysticial, line 77: During destruction of this class, `m_listeners` will be destructed before `m_download_thread` since destruction is in reverse order.

When `m_listeners` is destructed, the thread is still running and may try to call into it if a download finishes/fails at the same time.

So you should add a destructor for this class and have it explicitly end the thread as the very first thing. I suspect this might be partly responsible for issue you're having earlier where you need to delay destruction of the downloads to after the lock. (though I suspect there's a complicated deadlock going on with all these listeners)

Mysticial (approved):

PR #1310, opened 2026-06-29
Map detector revision

Developer-Butters:
Split out changes to PLZA Map detector from the helioptile fix.


Mysticial:
I have no idea how you managed to mess up the formatting that badly. So I just copied the two files verbatim from the other PR and merged it.

PR #1320, opened 2026-07-11
Unown RNG fixes for FRLG

theastrogoth:
This adds methods for Unown RNG manipulation, which are generated differently than other species. 

The Monean and Viapois chambers are also given a special approach to calibration in the Wild RNG program. 

Relies on https://github.com/PokemonAutomation/Packages/pull/75

### DIFF
--- a/SerialPrograms/Source/Pokemon/Pokemon_AdvRng.cpp
+++ b/SerialPrograms/Source/Pokemon/Pokemon_AdvRng.cpp
@@ -317,9 +317,9 @@ int slot_to_unownform(AdvEncounterSlot slot){
             return 24;
         case 'z':
             return 25;
-        case '?':
-            return 26;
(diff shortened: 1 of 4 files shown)


SerialPrograms/Source/Pokemon/Pokemon_AdvRng.cpp
```diff
@@ -28,7 +28,7 @@ void level_up_observed_pokemon(AdvObservedPokemon& pokemon, const StatReads& new
     pokemon.evs.emplace_back(new_evs);
 }
 
-uint32_t increment_internal_rng_state(uint32_t& state){
+uint32_t increment_internal_rng_state(const uint32_t& state){
```
Mysticial: Primitive types like this are better passed by value if they're not being modified.

PR #1322, opened 2026-07-12
Safari Auto Catch Improvements

Dalton-V:
Union pokedex registered detectors.
Move auto_catch to Navigation
Create Safari specific auto catch method using optimal safari actions.


SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
```diff
@@ -1193,6 +1196,278 @@ void heal_at_pokecenter(ConsoleHandle& console, ProControllerContext& context){
...
+    BlackScreenWatcher battle_end(COLOR_RED);
+    CatchFanfareDetector catch_detector(console.logger(), [&](float error_coefficient) -> bool{
+        catch_coefficient = error_coefficient;
+        return true;
+    });
+    BattleSelectionArrowWatcher nickname_question_arrow(
+        COLOR_RED,
+        &console.overlay(),
+        BattleConfirmationOption::YES
+    );
```
theastrogoth, line 1368: This might be useful in the non-safari auto catcher, too

SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
```diff
@@ -1193,6 +1196,278 @@ void heal_at_pokecenter(ConsoleHandle& console, ProControllerContext& context){
...
+                break;
+            case 1:
+                console.log("Party menu detected. Attempting to send out next Pokemon");
+                pbf_move_left_joystick(context, { 0, -1 }, 200ms, 300ms);
+                pbf_mash_button(context, BUTTON_A, 1000ms);
+                continue;
+            case 2:
+                console.log("Dex registration detected. Exiting battle...");
+                pbf_mash_button(context, BUTTON_B, 5000ms);
+                return catch_detected ? static_cast<int>(i) : 0;
```
theastrogoth: This is my own bad design -- I originally had this return as a failed catch if the fanfare was missed, even though the dex registration would imply a successful catch, since my detector would sometimes give false positives (on solid white screen transitions). 

Since the detector with the A button prompt is more reliable, this can probably also set `catch_detected = true;`

theastrogoth:
Looks good to me! I left a couple of small comments that mention possible improvements to my own code, but I can handle those separately if you'd rather not deal with them.

Mysticial (approved):
Rubber stamping since I've relinquished control of the FRLG to Astro and kichi.

Astro can merge when ready.

SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
Dalton-V: Added

SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
Dalton-V: I think I follow. Since we return right after I just updated it to `return 1;`. Let me know if that isn't what you had in mind.

Dalton-V:
Ope. I got the logic in auto catch crossed with safari in my update.. I didn't realize it was returning how many balls were thrown. Probably need to update the comment in the .h auto catch cause its wrong...

PR #1333, opened 2026-07-22
Make PaddleOCR a default resource and the default.

Mysticial:
Don't merge until we are ready to do this migration. This means the multi-line changes exceptions for Tesseract need to go in first.

Also: https://github.com/PokemonAutomation/Packages/pull/77

The renaming of the parameter is just to override everyone's configs which will already have been set to false.


jw098:
Replaced by #1335

PR #1338, opened 2026-07-24
fix BoxDraw so it doesn't crash if given improperly formatted string. Remove trailing zeros for FloatingPointWidget

jw098:


Common/Cpp/PrettyPrint.cpp
```diff
@@ -174,6 +174,18 @@ std::string tostr_fixed(double x, int precision){
     return std::format("{:.{}f}", x, precision);
 }
 
+std::string tostr_fixed_no_trailing_zero(double x, int precision){
+    std::string str = std::format("{:.{}f}", x, precision);
+    if (str.find('.') != std::string::npos) {
+        str.erase(str.find_last_not_of('0') + 1, std::string::npos);
```
Mysticial, line 180: Scientific notation isn't possible here right?

jw098: I believe so. I believe the F format doesn't use scientific notation. But the G format does.

Mysticial (approved):

PR #1339, opened 2026-07-24
add FRLG SeedsDatabase and simplify RNG program UI

theastrogoth:
Depends on https://github.com/PokemonAutomation/Packages/pull/78

This eliminates some UI fields (seed delay & nearby seeds) that were sources of common user errors by making use of seed lists.


Mysticial:
Rubber stamping since it's all FRLG.

PR #1353, opened 2026-08-02
move the colored text functions into their own file

jw098:


Common/Cpp/ColoredText.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifndef PokemonAutomation_ColoredText_H
+#define PokemonAutomation_ColoredText_H
+
+#include "Common/Cpp/Color.h"
+#include <string>
+
+namespace PokemonAutomation{
+
+inline size_t CURRENT_THEME = 0;
```
Mysticial: I guess this is not ideal, but the themes aren't visible here.

jw098: Yeah. I agree. Is there a better way?

Mysticial: Not really. Unless we move the theme definitions here.

Mysticial: Let's add an enum here:
```
enum class UiThemeMode{
    DEFAULT_MODE,
    DARK_MODE,
};
```
Then reuse this enum in the theme options.

Mysticial:
Merge conflicts.

PR #1357, opened 2026-08-07
RSE Updates

kichithewolf:
Detectors (~~not all are well~~ mostly tested):
- Load menu
- Party menu+slot
- Some Dialogs (incl. untested Pokenav)
- Start menu (no arrow/selection checking yet)
- Pokemon summary (1st page only)

New Programs:
- Ruby+Sapphire legendary reset
- Ruby+Sapphire gift reset

Program Updates:
- Legendary Run Away: added some movement configs
- Mew hunt: improved some movement timings, tested using a different device w/video output

To do:
- ~~RS Gift Reset~~
- E: Run Away needs Latis + Ray, need to account for PokeNav. This would be easier if Emerald's RNG issue is fixed, but that might be bad for RNG programs if/when we get RSE.
- Will handle rearranging programs, blocking off device selection, and updating the rest of the emerald stuff until after we know if/when the GBA ports come out and if we're getting all 3 at once.

If Emerald gets fixed:
- Starter reset will work, already tested
- Same with gift reset
- Same with legendary reset (but it may require some updates to start some encounters)
- Deoxys hunt can be hidden, soft reset will be faster
- Groudon/Kyogre can be removed from legendary run away as SR will be faster


SerialPrograms/Source/PokemonRSE/Inference/Dialogs/PokemonRSE_BattleDialogs.h
```diff
@@ -0,0 +1,115 @@
+/*  Battle Dialog Detector
+ *
+ *  From: https://github.com/PokemonAutomation/
+ *
+ */
+
+#ifndef PokemonAutomation_PokemonRSE_BattleDialogs_H
```
Mysticial, line 7: How similar are the RSE and FRLG battle menus? Since the gen3 games were built off the same-ish engine, maybe they are similar enough to share battle menu detectors? (among other things?)

kichithewolf: Sadly most of the UI elements are too different between games and languages to share detectors.

This one is closest, FRLG+Emerald, both non-Japan:
image
But if I attempt the same with Ruby, its way off:
image
And that's before adding Japan into the mix, both below are Ruby but the padding and positioning are different, and one (jpn) uses an arrow while the other (all other languages) uses a box as a selector.
image
Similar situation for Emerald:
image

I might actually have to adjust some of the menu positioning further for Ruby/Sapphire since like in FRLG some of the positioning is really tight. But that would have to wait for an official release so I know how much wiggle room there is.

Generally the UI Ruby/Sapphire is completely different while Emerald is much closer to FRLG. Which would line up with their release order.

Mysticial: oof

PR #1360, opened 2026-08-08
ImageViewRGB32: fix build if no backend defined

jw098:


Mysticial:
It will linker error even if we're not using it?

SerialPrograms/Source/CommonFramework/ImageTypes/ImageViewRGB32.cpp
```diff
@@ -52,10 +52,12 @@ bool ImageViewRGB32::save(const std::string& path) const{
 #endif
 #ifdef PA_IMAGE_BACKEND_OpenCV
     return OpenCV_save_image(*this, path);
-}
-
+#else
+    return false;
```
Mysticial, line 57: This won't work if `PA_IMAGE_BACKEND_Qt` is defined.

jw098: I just wanted a stub in case PA_IMAGE_BACKEND_None was defined. Do you prefer a separate if-def?

jw098:
> It will linker error even if we're not using it?

It linker errors when I try to define PA_IMAGE_BACKEND_None

Mysticial:
> > It will linker error even if we're not using it?
> 
> It linker errors when I try to define PA_IMAGE_BACKEND_None

Probably because it's being used somewhere. It's not going to work if you try to compile all of CC of course. But I'd expect it to work in the headless library since nothing there should be using images.

Mysticial (approved):

PR #1362, opened 2026-08-08
separate Switch_SettingsPanel from Switch_Settings.

jw098:


SerialPrograms/Source/NintendoSwitch/Controllers/NintendoSwitch_ControllerSettings.cpp
jw098, line 585: Because of this line, we have to pull in SettingsPanel, which isn't ideal.

Mysticial (approved):

PR #1365, opened 2026-08-08
FRLG summary reader fixes

theastrogoth:
This includes a few fixes to the FRLG summary reader:
- switches to only saving debug dumps on failed reads rather than on every read
- saves the full raw frame in addition to the crops and blurred+thresholded images
- removes the paths that used to be specific to Paddle, which would need multi-filtering added to be reliable. I'd never used/tested/maintained them, and I think the waterfill + template matching approach is consistent enough to rely on in all cases

The last point doesn't address Patrick's issues with PaddleOCR, since it didn't change the nature reader


Mysticial:
I see that you're yeeting some of the Tesseract/Paddle branching. Make sure both of them still work.

theastrogoth:
Confirmed working with both libraries (neither is used for digit reading now, since it's just template matching).

The Paddle branches weren't working for certain capture cards anyways, since they didn't have a multi-filter step

Mysticial (approved):

Mysticial:
Will merge when CI finishes green.

PR #1366, opened 2026-08-08
BDSP starter RNG manipulation

theastrogoth:
Blink detection, RNG state recovery, and a starter RNG program for BDSP.

[Lincoln's RNG docs](https://lincoln-lm.github.io/pokemon-rng-docs/) and Project_Xs were really helpful for getting started with state recovery

AI tools were used for troubleshooting and tidying up my ugly first drafts


SerialPrograms/Source/Pokemon/Pokemon_Xorshift128.h
```diff
@@ -0,0 +1,123 @@
+/*  Xorshift128
```
Mysticial, line 1: This different from Xoroshiro right? (which we already have for SwSh RNG)

Mysticial:
This one is definitely going in in multiple PRs. haha

SerialPrograms/Source/Pokemon/Pokemon_Xorshift128.h
theastrogoth: Yeah, the "ro" is for rotate, which doesn't happen in regular Xorshift

PR #1367, opened 2026-08-09
add intermediate dev mode

jw098:
I believe I moved all the internal stuff under INTERNAL_DEVELOPER_MODE. But please double check.
Also, I'm not sure if we want other stuff hidden under internal dev mode, such as RSE, ML stuff (although these are already available in the public repo).

### DIFF
--- a/SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.h
+++ b/SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.h
@@ -118,8 +118,8 @@ public:
 
     Pimpl<ErrorReportOption> ERROR_REPORTS;
 
-    BooleanCheckBoxOption DEVELOPER_MODE_CHECKBOX;
     StringOption DEVELOPER_TOKEN;
+    StringOption INTERNAL_DEVELOPER_TOKEN;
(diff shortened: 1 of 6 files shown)


Mysticial:
I'm thinking about also splitting the dev tokens here as well. Some only give regular dev access. Others, give both.

SerialPrograms/Source/CommonFramework/GlobalSettingsPanel.h
```diff
@@ -118,8 +118,8 @@ class GlobalSettings : public BatchOption, private ConfigOption::Listener, priva
 
     Pimpl<ErrorReportOption> ERROR_REPORTS;
 
-    BooleanCheckBoxOption DEVELOPER_MODE_CHECKBOX;
     StringOption DEVELOPER_TOKEN;
+    StringOption INTERNAL_DEVELOPER_TOKEN;
```
Mysticial: Oh I was thinking of keeping the same token field. But when we search the whitelist, we choose which flags to enable based on which whitelist it's in. So entering an internal-dev token will automatically enable both modes. Keeps the UI simpler.

PR #1369, opened 2026-08-09
PaddleOCR: add debugging, fix issue with dictionary loading to handle unicode paths

jw098:
-


jw098:
Not actually ready for review yet. But the CI runner doesn't work when it's a draft (for now).

jw098:
This is now ready for review.

PR #1375, opened 2026-08-11
Dont trigger CI on draft PR

Dalton-V:


Mysticial:
I'll let @pifopi review this one. lol

.github/workflows/cpp-ci-serial-programs-mac-intel.yml
```diff
@@ -3,6 +3,10 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
+    if: |
```
pifopi: if: github.event_name != 'pull_request' || github.event.pull_request.draft == false

sounds better if that's available ?

Dalton-V: whoops, good catch! I incrementally added these triggers when testing locally, should have stopped to think about cleaning it up after.

I verified that this still works on my local repos.

pifopi (approved):

PR #1378, opened 2026-08-11
add BDSP blink detection

theastrogoth:
Adds a few things related to blink detection:
- a way to set set the detection box and choose appropriate image templates depending on the scene (at Lake Verity, in the player character's bedroom, etc)
- a blink detector
   + converts template and capture to grayscale to mitigate in-game time of day differences or capture card hue differences
   + matches the template to the capture by searching across the detection box, which is big enough to allow for idle animations, and calculating cross-correlation
   + stores scores and timestamps in a vector to be used later to decide when blinks occurred (blinking increases the difference of the capture from the template for a few frames)
- a UI option for selecting one of the 8 character models (for template selection)

### DIFF
--- a/SerialPrograms/Source/PokemonBDSP/Inference/Rng/PokemonBDSP_BlinkScenes.cpp
+++ b/SerialPrograms/Source/PokemonBDSP/Inference/Rng/PokemonBDSP_BlinkScenes.cpp
@@ -1,0 +1,58 @@
+/*  BDSP Blink Scenes
+ *
+ *  From: https://github.com/PokemonAutomation/
+ *
+ */
+
+#include <string>
+#include "Common/Cpp/Exceptions.h"
+#include "PokemonBDSP/Options/PokemonBDSP_PlayerModelOption.h"
+#include "PokemonBDSP_BlinkScenes.h"
+
+namespace PokemonAutomation{
+namespace NintendoSwitch{
+namespace PokemonBDSP{
+
+
+std::vector<BdspEyeTemplate> lake_eye_templates(uint8_t player_model){
+    if (player_model < 1 || player_model > BDSP_PLAYER_MODEL_COUNT){
+        throw InternalProgramError(
+            nullptr, PA_CURRENT_FUNCTION,
+            "Unknown player model: " + std::to_string(player_model)
+        );
+    }
+    return {
+        BdspEyeTemplate{
+            "lake_templates/model" + std::to_string(player_model) + ".png",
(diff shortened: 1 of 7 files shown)


SerialPrograms/Source/PokemonBDSP/Inference/Rng/PokemonBDSP_EyeBlinkDetector.cpp
```diff
@@ -0,0 +1,176 @@
...
+        for (size_t x = 0; x < region_width; x++){
+            grey[y * region_width + x] = to_grey(region.pixel(x, y));
+        }
+    }
+
+    const double count = (double)(m_width * m_height);
+    double best = -1.0;
+
+    for (size_t dy = 0; dy + m_height <= region_height; dy++){
+        for (size_t dx = 0; dx + m_width <= region_width; dx++){
```
Mysticial: Can you leave a comment here explaining the math here? I see a variance calculation, though I can't tell how similar it is to the `image_stats()` function or if it duplicates it entirely.

SerialPrograms/Source/PokemonBDSP/Inference/Rng/PokemonBDSP_BlinkScenes.cpp
```diff
@@ -0,0 +1,58 @@
...
+std::vector<BdspEyeTemplate> lake_eye_templates(uint8_t player_model){
+    if (player_model < 1 || player_model > BDSP_PLAYER_MODEL_COUNT){
+        throw InternalProgramError(
+            nullptr, PA_CURRENT_FUNCTION,
+            "Unknown player model: " + std::to_string(player_model)
+        );
+    }
+    return {
+        BdspEyeTemplate{
+            "lake_templates/model" + std::to_string(player_model) + ".png",
```
Mysticial: I already merged the Packages PR for this, but can be change the naming convention to:
`LakeTemplates/Model-X.png`? (title case)

SerialPrograms/Source/PokemonBDSP/Inference/Rng/PokemonBDSP_EyeBlinkDetector.cpp
theastrogoth: I originally avoided `image_stats` to work in grayscale rather than RGB, but just using OpenCV's `matchTemplate()` instead of basically reimplementing it is probably the best option

PR #1380, opened 2026-08-12
upload standalone .exe file for windows builds

jw098:
- downgraded Qt version to match our official releases. So that someone can take the standalone .exe binary and test it themselves
- upload standalone .exe file for windows builds. I'm not sure if dropping in the standalone .exe file also works for Mac/Linux. If it does, then we could consider uploading standalone .exe files for them too.

### DIFF
--- a/.github/workflows/cpp-ci-serial-programs-base.yml
+++ b/.github/workflows/cpp-ci-serial-programs-base.yml
@@ -57,7 +57,7 @@ jobs:
       with:
         # Delete aqtsource once https://github.com/miurahr/aqtinstall/issues/1007 is fixed and released
         aqtsource: 'git+https://github.com/miurahr/aqtinstall'
-        version: '6.11.1'
+        version: '6.10.2'
         modules: 'qtmultimedia qtserialport'
 
     - name: Install dependencies (Ubuntu)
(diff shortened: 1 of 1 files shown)


.github/workflows/cpp-ci-serial-programs-base.yml
```diff
@@ -107,6 +107,13 @@ jobs:
         name: Serial Programs (os=${{inputs.os}} - compiler=${{inputs.compiler}})
         path: ${{env.UPLOAD_FOLDER}}
 
+    - name: Upload Executable Only
+      uses: actions/upload-artifact@v7
+      if: inputs.upload-build && startsWith(inputs.os, 'windows')
+      with:
+        name: Standalone Executable (os=${{inputs.os}})
+        path: ${{env.UPLOAD_FOLDER}}/SerialPrograms.exe
```
pifopi, line 115: any reason not to reuse the full folder with needed files, etc ?

pifopi (approved):

pifopi:
For mac and linux, I've tried in the past. Linux was working (=booting), mac I do not think so (but could not test). Not sure it's worth the pain with Koi's CI ?

.github/workflows/cpp-ci-serial-programs-base.yml
jw098: I don't understand the question. What are you suggesting?

Are you asking why I'm uploading the standalone .exe file, in addition to the full folder?

pifopi: Yes ! The exe file is included inside the folder that is already uploaded. And it has that and many more needed files (package, dll, etc) the only missing piece is dll for qt as they are very big and users might have a newer version locally!

jw098: My thought was that the standalone .exe is easier to download since it's smaller. For people who already have CC downloaded, they can just drop in the standalone SerialPrograms.exe to replace the old old.

pifopi: Works for me ! It's likely they'd need up to date package though but that can be downloaded manually on the side !

PR #1414, opened 2026-08-22
FRLG auto catch tweaks

theastrogoth:
Switches to using the same approach for exiting the battle for both `auto_catch()` and `auto_catch_safari()`. Moves a couple of things to shared helper functions. 

The only change to `auto_catch_safari()` behavior is an extra guard in case none of the detectors ever catch anything

### DIFF
--- a/SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
+++ b/SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
@@ -1196,22 +1196,102 @@ void heal_at_pokecenter(ConsoleHandle& console, ProControllerContext& context){
     }
 }
 
+// tries to pull up the start menu to check if currently in the overworld
+bool in_overworld(
+    ConsoleHandle& console,
+    ProControllerContext& context
+){
+    StartMenuWatcher start_menu;
+    context.wait_for_all_requests();
+    int ret = run_until<ProControllerContext>(
+        console, context,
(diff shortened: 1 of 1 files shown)


SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
```diff
@@ -1220,12 +1300,8 @@ int auto_catch(
             CatchFanfareDetector catch_detector(console.logger(), [&](float error_coefficient) -> bool{
                 catch_coefficient = error_coefficient;
                 return true;
-                });
-            BattleSelectionArrowWatcher nickname_question_arrow(
-                COLOR_RED,
-                &console.overlay(),
-                BattleConfirmationOption::YES
-            );
```
Dalton-V, line 1228: Is this not needed anymore? I think this is the detector you asked me to add to this method because I added it to the safari auto catcher.

Dalton-V (approved):

SerialPrograms/Source/PokemonFRLG/PokemonFRLG_Navigation.cpp
theastrogoth: I didn't think about the fact that it would also detect the "Send out next Pokemon?" selection box for a non-safari battle if something faints, so I took it out of the pre-catch loop there. 

It's still useful to watch for it while exiting, I think, and leaving it here means that the behavior for the safari catcher is unchanged.

PR #1424, opened 2026-08-25
FRLG Roamer: fix leave_pokecenter never actually exiting

adlopp:
travel_from_celio_to_kanto() does a single 1280ms hold left before calling leave_pokecenter(), assuming you can already move right after activate_roamer()'s dialogue ends. Turns out that's not reliable -- in testing (with Spanish game text) the character never even tries to turn left, it just falls straight into leave_pokecenter()'s blind walk-down and misses the door every time, eventually throwing the exit exception.

Turns out there's a leftover dialogue box still open after activate_roamer() finishes that eats the movement input. Waiting longer before moving doesn't help (tried up to 30s) since waiting alone doesn't close a dialogue box, but mashing A does close it -- which also confirms the game isn't actually frozen, just still showing text. Problem is mashing blind risks re-opening a new conversation with Celio if you overshoot.

So now it checks a screenshot for a dialogue box before moving, and only presses A once if it's actually there, re-checking in between presses. Also switched the left move to 5 separate steps instead of one long hold, since the continuous hold kept getting eaten somewhere.

None of this touches anything before the PID gets fixed in activate_roamer(), so it shouldn't affect the RNG result -- tested it and it's still landing on the right targets.

Heads up on the 220ms/step value for the movement: that number was reverse-calibrated from a single test session on my own hardware/capture setup (measured 400ms/step as ~9 tiles for 5 steps, scaled down to land on ~5). It's possible the dialogue fix alone was the actual bug and the original 1280ms hold would've been fine once unblocked -- discrete taps and a continuous hold don't necessarily cover distance at the same rate per ms, so this timing may need retuning on other hardware. Flagging this in case it needs adjustment rather than presenting it as a settled number.

Tested and working on FRLG (Spanish).

### DIFF
--- a/SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_RngNavigation.cpp
+++ b/SerialPrograms/Source/PokemonFRLG/Programs/RngManipulation/PokemonFRLG_RngNavigation.cpp
@@ -345,8 +345,42 @@ void hatch_daycare_egg(ConsoleHandle& console, ProControllerContext& context){
 
 
 void travel_from_celio_to_kanto(ConsoleHandle& console, ProControllerContext& context){
-    // assumes the player is standing on Celio's left (west) side
-    pbf_move_left_joystick(context, {-1, 0}, 1280ms, 300ms);
+    // assumes the player is standing on Celio's left (west) side.
+    //
+    // Confirmed by testing: the character never even turns left -- it goes
+    // straight into the "walk down" phase of leave_pokecenter(). Blind waiting
+    // (tried up to 30s) does not fix it, but blindly mashing A does have an
+    // effect (it re-opens Celio's dialogue), proving the game is responsive
+    // and a leftover dialogue box is the blocker, not an unresponsive game.
+    // So: check via screenshot whether a dialogue box is actually on screen,
+    // and only press A once if so, re-checking after each single press.
+    // Never press A when no dialogue is detected, to avoid re-triggering a
+    // fresh conversation with Celio.
+    {
+        WhiteDialogDetector dialog_detector(COLOR_RED);
+        bool cleared = false;
+        for (int attempts = 0; attempts < 10; attempts++){
+            context.wait_for_all_requests();
+            VideoSnapshot screen = console.video().snapshot();
+            if (!dialog_detector.detect(screen)){
+                cleared = true;
+                break;
+            }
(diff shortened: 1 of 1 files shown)


theastrogoth:
The problem with this fix is that, if the final dialogue box isn't closed, that means we're not generating the roaming legendary Pokemon at the target time, making it much harder (if not impossible) to actually do the intended RNG manipulation. This is why there wasn't already a mash B step to exit the dialogue, since that would hide any issues like this that come up

Are you sure it's still hitting the right targets? Have you gotten a shiny?

The underlying problem is likely that some languages (including Spanish) have an extra dialogue box compared to English. The better fix would be to account for it during the blind button press sequence. If you don't want to figure this out yourself, I'd be happy to take care of it if you can provide a video of all of Celio's dialogue

adlopp:
Yeah I got a shiny with the fix in, 4th automated attempt, so it's actually landing hits not just skipping the crash. Don't have the exact seed/advance numbers for that one though so can't say for sure it hit the exact PID

Can't get you a video unfortunately, my save's already past that point with Celio so the dialogue doesn't come up again for me, and I couldn't find existing footage clean enough either

Did find one thing that might help - an old guide for this method (manual, not language specific) says to watch for Celio's last text box specifically, "Thanks to you my dream came true" or whatever the exact line is, and time the final press off that instead of counting presses. Which lines up with what you're saying, a fixed count is gonna be fragile across languages no matter what number we land on. Might be worth having activate_roamer detect that last box with OCR instead

Can help if useful, or someone with a save before that point could probably get you the footage faster than me at this point

theastrogoth:
> Yeah I got a shiny with the fix in, 4th automated attempt, so it's actually landing hits not just skipping the crash. Don't have the exact seed/advance numbers for that one though so can't say for sure it hit the exact PID

Could you check the logs for what your in-game advances calibration was when you hit your target? If we're right about what is happening, your calibration should be somewhat negative.

> Did find one thing that might help - an old guide for this method (manual, not language specific) says to watch for Celio's last text box specifically, "Thanks to you my dream came true" or whatever the exact line is, and time the final press off that instead of counting presses. Which lines up with what you're saying, a fixed count is gonna be fragile across languages no matter what number we land on. Might be worth having activate_roamer detect that last box with OCR instead

Our FRLG programs do everything that is timing sensitive completely blind to make things as consistent as possible. My first attempts at RNG manipulation programs were much less reliable due to trying to detect a couple of things from the video capture. My preferred fix is to tweak the sequence as needed on a per-language basis

PR #1437, opened 2026-08-30
Refactor CMake to create GUI free library

jw098:
I also applied ifdef to all usages of Qt (in the files included in the Core library).


Mysticial:
Why is this touching so much? It's rather invasive too. If you're trying to make the minimal Qt-less library for Python integration, why is the entire media stack being pulled in? Same with the Discord integration?

jw098:
It's mostly because of GlobalSettingsPanel.cpp. It includes Discord stuff, ErrorReports, which in turn pull in media stuff.

Would you prefer if we just ifdef the stuff higher up in the chain?

NintendoSwitch_ProController, NintendoSwitch_Joycon, NintendoSwitch_PABotBase2_OemController, NintendoSwitch_PABotBase2_Controller all require "NintendoSwitch/NintendoSwitch_Settings.h", which eventually includes GlobalSettingsPanel. If we ifdef NintendoSwitch_Settings, that would solve a lot of the problems.

The other problem is that ControllerState includes OperationFailedException, which includes ProgramNotifications, and all the Discord stuff. So OperationFailedException would need to be ifdef as well.

Mysticial:
Yeah, ideally we break the dependencies much higher up.

`GlobalSettingsPanel` does seem to be the right place to if-def out all the options that aren't relevant to the Python lib.

`ControllerState` is a real problem as it exposes the fact that we don't have a "lightweight" exception class that doesn't have all the reporting stuff tied to it. I should take a look to see who catches those `OperationFailedException`s that are being thrown.

jw098:
Could ControllerState just throw a regular Exception (the base class for OperationFailed)?

EDIT: I decided to go with InternalProgramError, since I'm not sure that it should ever happen. Alternatively, maybe we could do UserSetupError?

SerialPrograms/Source/NintendoSwitch/Controllers/PABotBase2/NintendoSwitch_PABotBase2_OemController.cpp
```diff
@@ -308,12 +308,14 @@ bool PABotBase2_OemController::run_preconnect_configure(
         }
     }
 
+#ifdef QT_CORE_LIB
     NintendoSwitch::ControllerProfile profile =
```
Mysticial, line 312: Is there too much Qt to pull in the profile?

The profile isn't required, but if we ever get to the point where we initiate connections from Python, the controller will be completely black (more black than the default controller).

Mysticial (approved):

PR #1452, opened 2026-09-05
Aarch64 linux build

Gin890:
Let agent build on arm linux.
Add a reference build guide.


Common/Cpp/Hardware/Hardware_arm64_Linux.tpp
```diff
@@ -1,51 +1,106 @@
...
+#include <thread>
+#ifdef __APPLE__
+#include <sys/sysctl.h>
+#endif
+#include "Hardware.h"
+
+namespace PokemonAutomation{
+
+
+#ifdef __APPLE__
```
Mysticial: Instead of if-defing the entire file, can you just split this into 2 files. One for Mac and one for Linux.

Gin890: Followed Mysticial's comment

Mysticial (approved):

PR #1457, opened 2026-09-06
Update discord sdk

Gin890:
Also add a CMake function to ensure when the zip files are updated, CMake will extract the zip files automatically.


Mysticial (approved):

Mysticial:
rebase

PR #1499, opened 2026-09-21
Update max home boxes for Home October Update

Dalton-V:
Home is getting a max size increase from 6,000 to 9,000 in October when FRLG support is added. 

Not sure when you want to merge this in. It would cause issues if a user attempted to sort 300 boxes before the update. However, someone will probably complain that they can't sort all 300 boxes the day the update is released. 🤷


Mysticial:
This is a curious CI failure:
```
2026-09-21T02:30:05.9981536Z Starting: NintendoSwitch::UpdatePopupDetector - NintendoSwitch/UpdateMenuDetector/German_1_True.jpg
2026-09-21T02:30:06.0014198Z Use non-existent: 0000026B42A6EE30
2026-09-21T02:30:06.0337784Z Use non-existent: 0000026B42A6EE30
```
I don't think it has to do with your change. Might be my ongoing refactors. But I can't reproduce on master.

